PR #8, opened 2018-06-08
Feedback request for rtable.data.frame and rtable.groupby_df

dgkf:
> Sharing some exploratory work in userfacing API and dataframe/grouped_df integration. No intention to merge.

@waddella A few months back I did some tests in using this package in some of my own work. To try to move the user facing API towards a more declarative syntax, I wrote up some functions for building rtables from dataframes. 

For example, the `format` argument can accept a column name which contains format of the table for each row.

Additionally, there's handling of grouped dataframes which does automatic indenting by group.

Check out the `rtable.data.frame` and `rtable.grouped_df` functions and their examples for an idea of how this would work. There are still minor sticking points in the API that I'd like to improve, but as a whole I think it's much, much more approachable than constructing tables on a row-by-row basis. 

I know you've been recently considering more ease-of-use changes to the API and thought this early work might help add to that conversation.


waddella:
Hi @dgkf,

Thanks, I agree we need `data.frame` coercion function. I have now added `format_sprintf` (you are attributed), I think you have a more involved version of this.

I will look into it after my holidays.

Best,
Adrian

PR #21, opened 2019-10-25
Add alignment option

kieranjmartin:
A simple fix to allow the user to alter the alignment of the rtable, this is super useful; I haven't added documentation here (in fact the align input is obscured as its only associated with the rrow method); I suspect also it'd be easier to just let the user put in centre/right/left and add the text- part for them


danielinteractive:
@waddella @gmbecker I guess this can be closed given new version?

gmbecker:
Yes, the old version of rtables is no longer beign developed. @kieranjmartin thanks for your patch, soryr it didn't get incorporated. If you're still interested in rtables please check out the gabe_tabletree_work or devel branches.

PR #35, opened 2020-06-26
Work on documentation and toString

waddella:


waddella:
automation fails do to a bug from your branch. My changes should be OK.

PR #128, opened 2020-11-05
Create unit tests for formats and fix some

### LINKED ISSUE #6: Format of percentage not as expected
In master branch, a percentage format doesnt give .0 if a value rounds to 1/0.

E.G.
`   rcell(c(10, 0.1101), format = "xx (xx.x%)")`

returns
`10 (11%)`

rather than

`10 (11.0%)`

thomas-neitmann:
Close #6


thomas-neitmann:
@danielinteractive @gmbecker Please review and let me know if any changes are required 🙂

gmbecker:
The formatting code and specification is more @wadella's purview but we
will discuss it this week and get action on it either way.

Thanks @thomas-neitmann

FCACollin:
> Thanks @thomas-neitmann , leaving decision here to @gmbecker - maybe @FCACollin could be interested as he has written similar code before

@thomas-neitmann I think @danielinteractive refers to the following code. Please, check if there is any idea that could enhance your function.

```r
#' Formatting: XX as Formatting Function
#'
#' Translate a string where x and dots are interpreted as number place
#' holders, and others as formatting elements.
#'
#' @param str (`string`)\cr template.
#'
#' @export
#' @return A `rtables` formatting function.
#' @examples
#' test <- list(c(1.658, 0.5761), c(1e1, 785.6))
#'
#' z <- format_xx("xx (xx.x)")
#' sapply(test, z)
#'
#' z <- format_xx("xx.x - xx.x")
#' sapply(test, z)
#'
#' z <- format_xx("xx.x, incl. xx.x% NE")
#' sapply(test, z)
#'
format_xx <- function(str) {

  # Find position in the string.
  positions <- gregexpr(pattern = "x+\\.x+|x+", text = str, perl = TRUE)
  x_positions <- regmatches(x = str, m = positions)[[1]]

  # Roundings depends on the number of x behind [.].
  roundings <- lapply(
    X = x_positions,
    function(x) {
      y <- strsplit(split = "\\.", x = x)[[1]]
      rounding <- function(x)
        round(x, digits = ifelse(length(y) > 1, nchar(y[2]), 0))
      return(rounding)
    }
  )

  rtable_format <- function(x, output) {
    values <- Map(y = x, fun = roundings, function(y, fun) fun(y))
    regmatches(x = str, m = positions)[[1]] <- values
    return(str)
  }

  return(rtable_format)
}
```

thomas-neitmann:
Thanks @FCACollin! On first look I think we had basically the same idea: treat `xx`, `xx.`, `xx.xx` etc. as placeholders and substitute them with the actual values. However, in your code you are using `round` which will not necessarily give you the desired number of decimal places. Take `round(3, 0)` as an example which returns `3` whereas `sprintf("%.2f", 3)` returns `3.00` which is the desired outcome. In addition, for labels like `xx.x%` is it not enough to round the values but you have to multiply them by `100` to get the desired formatted string.

waddella:
@thomas-neitmann, apologies for the late reply. As @FCACollin mentioned this is a longer discussion which we are waiting with a decision on. The reason is that the format labels vocabulary is not complete yet and we plan to extend the "language", hence adding every format label added will be a constraint as we keep working on the format language. Therefore we do not plan to add regular expression based parsing for now.

That is unfortunate for all your good work as I will have to decline this PR. Please contact me in the future before you work on major pieces on `rtables` so we can discuss what the roadmap and whether your idea fits into the planned design changes.

Feel free to contact me if you want to discuss this in more detail.

waddella:
@thomas-neitmann, that said I will leave #6 open. The rounding is something that we will address. The reason for declining this PR is the extension of the format label language.

PR #141, opened 2020-12-04
Make `sort_at_path` return an empty table instead of failing.

danielinteractive:
Previously `sort_at_path` just failed with an error. I think it makes sense to just return an empty table immediately since nothing can be further sorted in an empty table.

Note: This failing currently blocks our internal development. So we should get this or another solution.


gmbecker:
Fixed (along with issue resulting in warning) and incorporated your test in above commit. Thanks for the contribution!

PR #215, opened 2021-07-16
internally proposed change around funargs

waddella:
@gmbecker please see the proposed change that addresses some issues we have in an internal project (`tern`).


waddella:
appologies, let me me fix that PR

PR #216, opened 2021-07-16
internally proposed change around funargs

waddella:
@gmbecker please have a look at the proposed change, the internal team has found this to solve some issues we get in `tern` (internal software) when preparing the switch to R 4.1.


waddella:
sorry I forgot to disable the circleci hooks, you can merge regardless

gmbecker:
@waddella I would still like more context around this change. What is the error? Because again, I develop rtables in 4.1.0 and things are passing/working/etc for me, so the actual problem is unlikely (not impossible, but unlikely) to be the change between 3.6.x and 4.1.0

waddella:
@gmbecker this is the error I get (in an internal package) without the change

image

I have sent you instructions per email on how to replicate the error.

waddella:
see #226

PR #271, opened 2021-11-18
use em dash for separators in ASCII

### LINKED ISSUE #270: use em dash instead of hyphen
https://en.wikipedia.org/wiki/Wikipedia:Hyphens_and_dashes

The em-dash will remove the gaps in the lines

```
ABC            setosa   versicolor   virginica
----------------------------------------------
Sepal.Length                                  
  mean         5.006      5.936        6.588  
Sepal.Width                                   
  mean         3.428       2.77        2.974  
```

vs

```
ABC            setosa   versicolor   virginica
——————————————————————————————————————————————
Sepal.Length                                  
  mean         5.006      5.936        6.588  
Sepal.Width                                   
  mean         3.428       2.77        2.974  
```

waddella:
closes #270

just when I got used to `---` as separators :smiley:

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -34,7 +34,7 @@ License: Apache License 2.0 | file LICENSE
 Encoding: UTF-8
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.1.1
+RoxygenNote: 7.1.2
 VignetteBuilder: knitr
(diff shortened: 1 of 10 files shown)


gmbecker:
I don't object to using em-dash here (though I also don't really see the big deal about it as the dashed line header seems fine to me as well) but this is causing test failures. Can you please update the failing tests @waddella

PR #275, opened 2021-12-06
Hotfix as license missing CPH 

epijim:
Ross noticed that we hadn't filled in the templates in the license


gmbecker (approved):

gmbecker:
Thank you for this PR. Due to issues with our automation I went ahead and directly made the relevant fixes but we really appreciate both the report and you putting together the PR

PR #300, opened 2022-02-24
add variable labels related functions back but this time deprecated

waddella:
@gmbecker, we have moved these functions to a different currently internal package. However we need to leave them in `rtables` in a deprecated state for a couple releases. Please check the deprecation message.

### DIFF
--- a/R/deprecated.R
+++ b/R/deprecated.R
@@ -1,4 +1,9 @@
 #nocov start
+
+deprecation_message <- function() {
+    message(paste("Note that the function", as.character(sys.call()), "is deprecated."))
+}
+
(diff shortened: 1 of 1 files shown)


gmbecker:
@waddella Why? They are in a package which is a hard dependency for rtables, and are exported there, so they will be available for any users of rtables and software built on rtables. The only thing this would break is people calling `rtables::with_label` which they really shouldn't be doing (for this exact reason, among others).

They are moved so that they available to both rtables and rlistings. 

Having them exported by rtables and one of its dependencies at the same time is wrong.

waddella:
you are right the label functions are in `formatable`, we should deprecate them in `formatable` eventually.

PR #352, opened 2022-08-16
Merge pre-release into main

cicdguy:
Removes old files


gmbecker:
@cicdguy why is this failing the roxygen check when main doesn't and no .R files are changed by the PR???

gmbecker:
The issue is that the Rd that is in main now is correct, and the updated thing you get by running roxygen again is wrong.

Oh, I know why I think. Its an importParams, so roxygen is actually hitting the wrong version of formatters somehow. (or the relevant fix never got ported over to pre-release for formatters?)

IF you look at the changes there you see `" @details"` in the Rd file, which isn't what we want. And is due to roxygen being too dumb to understand that that space doesn't mean anything...

cicdguy:
> or the relevant fix never got ported over to pre-release for formatters?

Let me port that and check again.

cicdguy:
Yep - that was it. Looks right now.

gmbecker (approved):
Looks good now. Thanks

cicdguy:
Feel free to merge when you can. Turns out I'm not authorized to merge

Screen Shot 2022-08-16 at 7 46 02 PM

gmbecker:
oh. thats weird.

PR #359, opened 2022-08-29
353 refactor get formatted cells@main

### LINKED ISSUE #353: refactor `get_formatted_cells`, `RowTable` method to work off CellValue objects directly
@melkiades Please take a look at this and consider it for a first issue.

Currently, the `RowTable` method for `get_formatted_cells` uses `row_values` to extract the *raw* values from the row's cells, and then later passes them, along with a format and na_str to `format_rcell`. This works but causes annoying regressions whenever we (very occasionally) want to add more information that the CellValue caries around.

`format_rcell`, however, is intended to be able to work directly on the `CellValue` (S3 class for performance reasons, see it's constructor for its details) objects carried around by a table row.

Refactor (and thus simplify) this method so it operates on the CellValue objects.

Note: this will likely involve narrow changes to `format_rcell`, this is ok, but make sure they are backwards compatible.

Melkiades:
Fixes #353


R/tt_toString.R
```diff
@@ -590,29 +590,11 @@ setMethod("get_formatted_cells", "ElementaryTable",
...
-
-            matrix(unlist(Map(function(val, format, spn, na_str) {
+            matrix(unlist(Map(function(val, spn) {
                 stopifnot(is(spn, "integer"))
-                val <- if(shell) format else paste(format_rcell(val, format, na_str = na_str),
-                                                   collapse = ", ")
+                val <- paste(format_rcell(val, shell = shell), collapse = ", ")
                 rep(list(val), spn)
-            }, val = row_values(obj), format = format, spn = row_cspans(obj), na_str = na_str)),
+            }, val = row_cells(obj), spn = row_cspans(obj))),
```
gmbecker, line 597: The key here (and the reason the tests didn't pass, I suspect), is that this doesn't preserve the inheritance from the row object if the cell object doesnt have a format or an na_str value. 

We need a way to make sure that still happens, even though we're operating on the cells now.

Perhaps accept a `parent_row` argument, which defaults to NULL, but which when not null, provides format and na_str information if and only if the corresponding value on the cell is not set.

tests/testthat/test-formatting.R
```diff
@@ -112,3 +112,137 @@ test_that("format_na_str functionality works in get_formatted_cells (ie printing
...
+test_that("Test matrix obtained from get_formatted_cells with only specific na values", {
+
+  # Introducing NAs
+  iris2 <- iris
+  iris2$Sepal.Length[sample(seq_len(nrow(iris2)), size = 10)] <- NA
+  iris2$Petal.Width[sample(seq_len(nrow(iris2)), size = 10)] <- NA
+
+  # Custom summary function to add specific formats for na values to be reflected
+  s_summary <- function(x) {
+    stopifnot(is.numeric(x))
```
gmbecker, line 210: No real reason to have the stopifnot here as this is test code.

tests/testthat/test-formatting.R
```diff
@@ -112,3 +112,137 @@ test_that("format_na_str functionality works in get_formatted_cells (ie printing
...
+    )
+  }
+
+  afun <- make_afun(
+    fun = s_summary,
+    .formats = c(mean = "xx.xx"),
+    .labels = c(mean = "Mean")
+  )
+
+  afun_isridiculous <- make_afun(afun,
```
gmbecker, line 222: Typically we don't use make_afun in tests unless we are _testing `make_afun` itself_, as it embodies a (very large) amount of additional complexity that is (usually) not relevant for what we're trying to test.

That said, we do also need to maintain the tests for `make_afun` so that coverage is good.

R/tt_toString.R
gmbecker, line 597: Looks like this is missing `shell` which you made an argument to `format_rcell`. Not sure about that change, shell seems conceptually to fit better at the `get_formatted_cells` level, but lets talk about it in the meeting. If if we will do make the change, it will need to be passed down in the `mapply`.

R/tt_toString.R
```diff
@@ -590,29 +590,11 @@ setMethod("get_formatted_cells", "ElementaryTable",
 #' @rdname gfc
 setMethod("get_formatted_cells", "TableRow",
           function(obj, shell = FALSE) {
-            default_format <- if (is.null(obj_format(obj))) "xx" else obj_format(obj)
-            format <- lapply(row_cells(obj), function(x) {
```
gmbecker, line 594: Note the obj here is the _row_ object, NOT the cell object, important to preserve that behavior

gmbecker (changes requested):
Please make sure the changes pass check (ie passing all existing tests) in addition to reading and addressing specific comments

tests/testthat/test-formatting.R
Melkiades: I did use that to test that the na_str was propagated by the CellValues object instead of the previous way (through `get_formatted_cells`)

R/tt_toString.R
Melkiades: ahhhh, I see! I took a moment, but now I understand. Also `TableRow` could have a format definition... mmmh we need to talk about format hierarchies because it is sometimes hard to understand which format will be finally picked and from where

R/tt_toString.R
Melkiades: actually, the error was coming from the fact I was not considering the option by which format is a function. I think I found a way to get rid of `shell` in `format_rcells`

tests/testthat/test-subset-access.R
```diff
@@ -83,8 +83,8 @@ test_colpaths <- function(tt) {
     cdf2 <- make_col_df(tt, visible_only = FALSE)
     res3 <- lapply(cdf$path, function(pth) rtables:::subset_cols(tt, pth))
     res4 <- lapply(cdf2$path, function(pth) rtables:::subset_cols(tt, pth))
-    expect_identical(res3, res4[!is.na(cdf2$abs_pos)])
-    expect_identical(res3, lapply(seq_len(ncol(tt)),
+    testthat::expect_identical(res3, res4[!is.na(cdf2$abs_pos)])
```
gmbecker, line 86: There's no need to use `::` here (or nearly anywhere in package code), so please remove it here and elsewhere.

tests/testthat/test-formatting.R
```diff
@@ -1,6 +1,6 @@
+library(dplyr)
```
gmbecker, line 1: If this is required it should occur once in the setup-file unless there is a reason to localize it (which in this and virtually all cases, there  isn't)

tests/testthat/test-formatting.R
```diff
@@ -112,3 +112,136 @@ test_that("format_na_str functionality works in get_formatted_cells (ie printing
     expect_identical(get_formatted_cells(tbl)[3, 1, drop = TRUE],
                      "Ridiculous")
 })
+
+test_that("Test matrix obtained from get_formatted_cells as is into ASCII format", {
+
+  # Table builder with rtables
+  tbl <- basic_table() %>%
```
gmbecker, line 119: We can be more efficient with our tests here. 

If we use an afun such as

```
weird_afun <- function(x, ...) {
in_rows(cell_fmt = rcell(mean(x, na.rm = TRUE), format = "xx.x"),
              no_cell_fmt = rcell(median(x, na.rm = TRUE)),
              cell_na_str = rcell(NA, format = "xx.x", na_str = "haha"),
              no_cell_fmt = rcell(NA))
}
```

And then do 

```
DM2 <- DM
DM2$AGE[1] <- NA
tbl <- basic_table() %>%
    split_cols_by("ARM") %>%
    analyze("AGE", weird_afun, format = "xx.xx", na_str = "woah") %>%
    build_table(DM2)
``` 

Then we can check all four cases against that single matrix_form.

R/tt_toString.R
```diff
@@ -590,29 +590,30 @@ setMethod("get_formatted_cells", "ElementaryTable",
...
+            pr_row_na_str <- obj_na_str(obj) %||% "NA"
+
+            matrix(unlist(Map(function(val, spn) {
                 stopifnot(is(spn, "integer"))
-                val <- if(shell) format else paste(format_rcell(val, format, na_str = na_str),
-                                                   collapse = ", ")
+
+                # Case it is shell
+                if (shell) {
+                    format <- obj_format(val)
```
gmbecker, line 603: This may be a good argugment for actually pushing `shell` behavior down to the `format_rcell` level. 

That or having `resolve_format(cell, row)` and `resolve_na_str(cell, row)` functions to call here, as I'm not a fan of this logic appearing in 2 places.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 ## rtables 0.5.2.1
  * dev version increment
```
gmbecker, line 2: package versions are defined in our repo by the first commit to bear that version number in `DESCRIPTION` therefor this isn't really correct. What you have there, if we put it in the news file, will be an entry under `0.5.2.2`, though its an implementation detail so I'm not sure it needs to go in the NEWS file to begin with. That is not to say it wasn't valuable, just not something someone reading a NEWS file is likely to care about

gmbecker (changes requested):

tests/testthat/test-subset-access.R
Melkiades: I added it only because if the package is not loaded, `lintr` complains. Removing it now. There are many other cases where `::` and `:::` are present (the latter should not pass the `--as-cran` tests, is the automation using it?). Should I remove them all? It could be better to add an issue for this

NEWS.md
Melkiades: it makes sense. `dev version increment` is a placeholder?

R/tt_toString.R
Melkiades: yes, pushing `shell` down was more elegant, but as you said less useful to have such a variable for `format_rcell`. I will think what is best imo and propose it to you here

tests/testthat/test-formatting.R
Melkiades:
```
Error in analyze(., "AGE", weird_afun, format = "xx.xx", na_str = "woah") :
unused argument (na_str = "woah")
```
Also calling the docs this parameter is not present. ps: I tried to find this functionality since the beginning btw, that is why I used `make_afun` in the first place. Also tried unsuccessfully: `extra_args = list(na_str = "woah")`

R/tt_toString.R
Melkiades: The current solution looks to me like the best. Still, I had no idea that format can have all these possible values and behaviors. Something showing this schematically is needed imo.

tests/testthat/test-formatting.R
Melkiades: I adapted the test but still no solution for the above comment

gmbecker: Yes, na_str is less pervasive than I had thought. You can currently only set it within in_rows (which is what `make_afun` does, so that won't be any different than the "haha" case above). Neither Table nor Row based classes currently have a slot for it, only `CellValue` and `RowsVerticalSection`.

For now leave that off of the test for na_str inheritence, and we will add it back in when we implement that functionality more widely.

tests/testthat/test-subset-access.R
gmbecker: Dont worry about that, i'll take care of it. The rule is we need `::` whenever we are utilizing a suggests package, rather than a hard dependency (and those calls must come after a call to `requireNamespace(<suggested pkg>)`)

Currently, that is `flextable`, used (unsurprisingly) in `tt_to_flextable`. Everywhere else its not necessary and my next commit will remove it there.

tests/testthat/test-formatting.R
Melkiades: ok perfect!

tests/testthat/test-subset-access.R
Melkiades: Ok, I see, I added it as a rule ;)

tests/testthat/test-formatting.R
```diff
@@ -112,3 +111,67 @@ test_that("format_na_str functionality works in get_formatted_cells (ie printing
...
+        build_table(DM2)
+    
+    # Get the ASCII table
+    result <- get_formatted_cells(tbl) # Main function
+    
+    # Expected data-set is built with dplyr
+    expected <- DM2 %>% 
+        dplyr::group_by(ARM) %>%
+        summarise(
+          m_age = deci_format(mean(AGE, na.rm = TRUE), dig = 1),
```
gmbecker, line 152: please use format_value (from formatters) instead of deci_format here, as the logic is similar but not identical (one of the reasons we had to write the formatting code in the first place)

gmbecker (changes requested):
Getting close, just a couple of minor changes.

PR #394, opened 2022-09-12
fixing cat and print outputs in tests

### LINKED ISSUE #368: Removing prints from tests
This enhancement focuses on the readability of tests and involves only a couple of functions that must maintain the ability to give meaningful errors.

Melkiades:
I added a flag for `row/col_summary` called `verbose`, which may be `TRUE` as default (maybe it is better, atm it is `FALSE`) because this function is always used to assign a variable that is then tested. For coverage, I captured the case in which it is `TRUE`. I think this is not the best behavior (printing out with `cat` as a standard). I also fixed some months ago the same problem in [tern#test-summarize_patients_exposure_in_cols.R#L107](https://github.com/insightsengineering/tern/blob/673aaca40ae772584ed4c2f17f779918f96ca669/tests/testthat/test-summarize_patients_exposure_in_cols.R#L107)  with the promise to bring this up in the future.

Fixes #368

### DIFF
--- a/R/summary.R
+++ b/R/summary.R
@@ -94,7 +94,6 @@ row_paths_summary <- function(x) {
   )
 
   txt <- mat_as_string(mat)
-  
   cat(txt)
   cat("\n")
 
@@ -121,7 +120,6 @@ col_paths_summary <- function(x) {
   )
 
   txt <- mat_as_string(mat)
-  
   cat(txt)
   cat("\n")
 
--- a/R/summary.R
+++ b/R/summary.R
@@ -5,8 +5,6 @@
 #' @title Return List with Table Row/Col Paths
 #'
 #' @param x an rtable object
-#' @param verbose (`logical`) \cr Defaults to `FALSE` and determine wether the
-#'   summary should be printed to console. 
 #'
 #' @export
 #' @return a list of paths to each row/column within \code{x}
@@ -40,6 +38,8 @@ col_paths <- function(x) {
 #' Print Row/Col Paths Summary
 #'
 #' @param x an rtable object
+#' @param verbose (`logical`) \cr Defaults to `TRUE` and determine whether the
+#'   summary should be printed to console. 
 #'
 #' @export
 #' @return A data.frame summarizing the row- or column-structure of \code{x}.
@@ -73,7 +73,7 @@ col_paths <- function(x) {
 #'    rrow("r1", 1, 2, 1, 2), rrow("r2", 3, 4, 2,1)
 #' )
 #' col_paths_summary(tbl3)
-row_paths_summary <- function(x, verbose = FALSE) {
+row_paths_summary <- function(x, verbose = TRUE) {
   stopifnot(is_rtable(x))
   stopifnot(is.logical(verbose))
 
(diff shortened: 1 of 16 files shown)


tests/testthat/test-regressions.R
```diff
@@ -16,7 +16,7 @@ test_that("manually created label l rows are always visible",
 ## was error before rtables 0.3.2.16
 test_that("printing table with 0 rows works", {
     norows <- rtable(c("hi", "lo"))
-    expect_identical(print(norows), norows)
+    expect_identical(capture.output(print(norows)), capture.output(norows))
```
gmbecker, line 19: this isn't teh same test, please do the same as elsewhere with `capture.output({prout <- print(tbl)})` ...

tests/testthat/test-printing.R
```diff
@@ -16,7 +16,7 @@ test_that("toString method works correclty", {
         }) %>%
         build_table(iris, hsep = "=")
 
-    print(tbl)
+    invisible(capture.output(print(tbl)))
```
gmbecker, line 19: I don't think we need invisible here, just capture.output. Same below.

R/summary.R
```diff
@@ -73,8 +73,9 @@ col_paths <- function(x) {
 #'    rrow("r1", 1, 2, 1, 2), rrow("r2", 3, 4, 2,1)
 #' )
 #' col_paths_summary(tbl3)
-row_paths_summary <- function(x) {
+row_paths_summary <- function(x, verbose = TRUE) {
```
gmbecker, line 76: I'd like to separate this into a separate issue, rather than covering it here because I feel this change invovles a larger discussion about what we want summaries to do and how many of them we have, as I mentioned in our standup

gmbecker (changes requested):
Please split out the change to the summary functions and just use capture.output in the relevant tests for now.

R/summary.R
Melkiades: Reverted and opened a separated issue #395

R/summary.R
```diff
@@ -1,5 +1,3 @@
-
-
```
gmbecker, line 2: Please remove all whitespace only changes

tests/testthat/test-formatting.R
```diff
@@ -37,7 +37,7 @@ test_that("table_shell works", {
                   rrow("Hiya", "xx (xx.x%)", "xx.x - xx.x"),
                   rrow("bye", "xx.xx", "xx.xx"))
 
-    expect_identical(toString(tblsh),
+    expect_identical(toString(tblsh),
```
gmbecker, line 40: whitespace change

gmbecker (changes requested):
Please remove all whitespace only changes and then this will be ready to merge.

tests/testthat/test-formatting.R
Melkiades: yes, I usually run `styler:::style_selection()` or `styler:::style_active_pkg()` to avoid checking every whitespace manually, but it seems you do not have a general coding standard adopted, or at least I do not know which one. If I run the above commands it is using `tidyverse` standard and it is changing basically all the code. For example, my Rstudio-server does not recognize the double spaces as single tabs, so I have to double space every line manually which is quite annoying and prone to errors. I know you can set up different standards but this would not be in line with `tern` and all the other packages that use `styler`. I suggest adopting the above general standard, also to improve readability. We can discuss this later today. For the moment, of course, I will fix these two white spaces

gmbecker (approved):

Issue #362, opened 2022-08-30
Fix pkgdown reference index


waddella:
It seems to be broken at the moment

gmbecker:
https://github.com/r-lib/pkgdown/issues/2192

PR #406, opened 2022-10-05
Add docs workflow

cicdguy:
The following needs to be remediated:

```
  All topics must be included in reference index
✖ Missing topics: table_inset,VTableTree-method
ℹ Either add to _pkgdown.yml or use @keyword internal
```

Closes https://github.com/Roche/rtables/issues/363


Melkiades:
I think the failings are due to #362 and the related issue in `pkgdown`. Did this ever happen to you @cicdguy?

PR #413, opened 2022-10-07
410 new tests inset@main

### LINKED ISSUE #410: Write test of inset+pagination functionalities that includes referential footnotes.

Please briefly describe your problem and, when relevant, the output you expect.
Please also provide the output of `utils::sessionInfo()` or
`devtools::session_info()` at the end of your post.

If at all possible, please include a [minimal, reproducible
example](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example).
The `rtables` team will be much more likely to resolve your issue if they are
able to reproduce it themselves locally.

Please delete this preamble after you have read it.

---

your brief description of the problem

```r
library(rtables)

# your reproducible example here

```

Melkiades:
Fixes #410


Melkiades:
The error regarded the hsep that was set to a different default. Now it is specified and it should work fine

PR #426, opened 2022-10-20
Use the flatly bootswatch theme

cicdguy:
Instead of cerulean. Something's off with cerulean.
Closes https://github.com/Roche/rtables/issues/425

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -1,16 +1,14 @@
-
-home:
-  strip_header: true
+url: https://roche.github.io/rtables
 
 template:
   bootstrap: 5
-  bootswatch: cerulean
+  bootswatch: flatly
 
 navbar:
   type: default
   right:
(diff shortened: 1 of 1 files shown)


_pkgdown.yml
```diff
@@ -1,16 +1,14 @@
...
   bootstrap: 5
-  bootswatch: cerulean
+  bootswatch: flatly
 
 navbar:
   type: default
   right:
-  - icon: fa-github fa-lg
-    href: https://github.com/roche/rtables
+    - icon: fa-github fa-lg
```
gmbecker, line 10: @cicdguy are these whitespace changes necessary?

gmbecker (changes requested):

_pkgdown.yml
gmbecker: Ok, works for me

gmbecker (approved):

PR #442, opened 2022-11-09
Devel

StefanThoma:


Melkiades:
I do not know what you did but please retry by branching from an UPDATED main and adding your modifications from there. Also, before that, could you write an issue with a short description of the feature you want to implement? Then branches should be named `111_description@main` where the first number (here 111) is the number of the issue it refers to and the description is a title for the PR.

PR #457, opened 2022-11-15
tests for cell and label wrapping

### LINKED ISSUE #450: construct tests for new cell and label wrapping
set colwidths (`paginate_table(, colwidths = .)`, `toString(, widths = .)`) and ensure both printing and pagination work.

You'll need the latest from main for both `rtables` and `formatters`

Melkiades:
Fixes #450

### DIFF
--- a/tests/testthat/setup-fakedata.R
+++ b/tests/testthat/setup-fakedata.R
@@ -174,11 +174,11 @@ tt_to_test_wrapping <- function() {
 
 tt_for_wrap <- tt_to_test_wrapping()
 
-# Helper function in R base to count how many times a pattern appears in a string
-.count_str_for_tests <- function(str_test, pat, negate = FALSE) {
+# Helper function in R base to count how many times a character appears in a string
+.count_chr_from_str <- function(str, chr, negate = FALSE) {
     if (negate) {
-        nchar(gsub(pat, "", str_test, fixed = TRUE))
+        nchar(gsub(chr, "", str, fixed = TRUE))
     } else {
-        nchar(str_test) - nchar(gsub(pat, "", str_test, fixed = TRUE))
+        nchar(str) - nchar(gsub(chr, "", str, fixed = TRUE))
     }
 }
--- a/tests/testthat/setup-fakedata.R
+++ b/tests/testthat/setup-fakedata.R
@@ -174,7 +174,8 @@ tt_to_test_wrapping <- function() {
 
 tt_for_wrap <- tt_to_test_wrapping()
 
-# Helper function in R base to count how many times a character appears in a string
+# Helper function in R base to count how many times a character appears in a string.
+# W: this works only for counting single characters from a single string of txt
 .count_chr_from_str <- function(str, chr, negate = FALSE) {
     if (negate) {
         nchar(gsub(chr, "", str, fixed = TRUE))
--- a/tests/testthat/setup-fakedata.R
+++ b/tests/testthat/setup-fakedata.R
@@ -152,3 +152,33 @@ export_fact <- function() {
 }
 
 tt_to_export <- export_fact()
+
+# Creating data-set with wide content to test wrapping
+tt_to_test_wrapping <- function() {
+    trimmed_data <- ex_adsl %>% 
+        filter(SEX %in% c("M", "F")) %>% 
+        filter(RACE %in% levels(RACE)[1:2]) 
+    
+    levels(trimmed_data$ARM)[1] <- "Incredibly long column name to be wrapped"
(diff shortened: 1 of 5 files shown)


tests/testthat/test-printing.R
```diff
@@ -328,8 +330,66 @@ test_that("Inset works for table, ref_footnotes, and main footer", {
...
+                  "AMERICAN            ")
+    expect_identical(splitted_res2[8:10], expected)
+
+    # Test if it works with numeric values
+    tt_simple <- basic_table() %>%
+        analyze("AGE", format = "xx.xxxx") %>%
+        build_table(ex_adsl)
+    result <- toString(matrix_form(tt_simple), widths = c(2, 3), col_gap = 1)
+    sre3 <- strsplit(result, "\n")[[1]]
+    expected <- c("   all", "——————", "   obs", "Me 34.", "an 88 ")
```
gmbecker, line 389: This is testing the buggy behavior. I'll be pushing a fix to that shortly. You'll need to rebuild the test after that goes in so it enforces the right behavior.

gmbecker (changes requested):

tests/testthat/test-printing.R
Melkiades: done

tests/testthat/setup-fakedata.R
```diff
@@ -152,3 +152,33 @@ export_fact <- function() {
...
+        split_rows_by("RACE", split_fun = drop_split_levels) %>%
+        analyze(c("AGE", "EOSDY"), 
+                na_str = "A very long content to_be_wrapped_and_splitted",
+                inclNAs = TRUE) %>%
+        build_table(trimmed_data)
+}
+
+tt_for_wrap <- tt_to_test_wrapping()
+
+# Helper function in R base to count how many times a pattern appears in a string
```
gmbecker, line 177: This only works if pattern is a single character, please change the name of it and document that fact in a comment, in case we ever want to use it elsewhere

gmbecker (changes requested):

tests/testthat/test-printing.R
```diff
@@ -146,8 +146,12 @@ test_that("newline in column names and possibly cell values work", {
...
     expect_identical(dim(matform2$strings),
                      c(18L, 5L))
-    expect_identical(attr(matform2, "nlines_header"),
-                     4L)
+    # expect_identical(attr(matform2, "nlines_header"),
+    #                  4L)
+    expect_identical(attr(matform2, "nrow_header"),
+                     3L)
+    expect_identical(attr(matform2, "ncols"),
+                     4)
```
Melkiades, line 154: @gmbecker, the commented line is failing as there is not such attribute anymore. I think the name of it was changed to `nrow_header`, but its associated value is `3L` and not `4L`. So, just to show, I added the variable that has 4 as value (not `4L` btw). Checking the tested matrix though, I do not know what is the number of rows we expect from the header (I would say four??). Here, it is:
```
Ethnicity               Arm                 Arm       
Factor2                    1                   2        
                    Male     Female     Male     Female 
                   (N=256)   (N=248)   (N=248)   (N=248)
————————————————————————————————————————————————————————
Caucasian                                               
  Level A                                               
    mean            56.42     55.57     56.19     54.53 
    median          55.91     55.42     58.40     51.73 
  Level B                                               
    mean            54.28     55.24     54.22     54.92 
    median          54.71     55.47     54.96     55.17 
African American                                        
  Level A                                               
    mean            55.77     55.33     56.26     54.30 
    median          55.06     54.39     57.20     53.94 
  Level B                                               
    mean            53.85     56.55     56.80     55.88 
    median          54.36     57.21     57.39     55.71 
```

tests/testthat/setup-fakedata.R
Melkiades: Done. Let me know if it is fine as it is now ;)

tests/testthat/test-printing.R
Melkiades: Unrelated and maybe not an issue at all but I noticed an indentation on the first `topleft` name and not on the second. As I do not see any indentation directive in the test, is it a designed standard? Here a screenshot of it printed to console:
Screenshot 2022-11-17 163212

gmbecker: nrow header and nlines_header have always been different. 

It is true that the attribute no longer exists, but the getter still does, and should have been used anyway (`mf_nlheader(matform)`)

In fact we need to search out all calls to attr and get rid of most of them as they penetrate the abstraction

Melkiades: Using search it seems like this was the only case where `attr(` is used, so no problem!

gmbecker (approved):

PR #460, opened 2022-11-17
Corrected "devtool" in readme

dmenne:
It should be "devtools", but "remotes" is the primary source anyway


gmbecker:
This fix has been bundled with another PR, thank you much for the report!

PR #473, opened 2022-11-25
Subsetting should lose header and footer if not differently specified

### LINKED ISSUE #456: Why subsetting loses title and subtitles when there are columns while not the footers? Is it expected?
This a rare case, I think but I encountered it while testing specific parts of the table for which I used a lot of subsetting. To reproduce it run the following:
```r
library(rtables)
tbl <- basic_table(title = "a", subtitles = "b", main_footer = "c", prov_footer = "d") %>%
    split_rows_by("SEX") %>% 
    split_cols_by("ARM") %>%  # if this is omitted title and subtitles work
    analyze("AGE") %>%
    build_table(ex_adsl)

cat(toString(matrix_form(tbl[2,1], TRUE)))
```
When fixed, this needs a change in tests like for #457

Melkiades:
Fixes #456


R/tt_pos_and_access.R
```diff
@@ -939,42 +990,49 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
-       length(i) == 1L &&
-       drop) {
+                         keep_titles = keep_titles,
+                         keep_fnotes = keep_fnotes)
+
+    # Dropping everything
+    if (drop) {
+        if (!(length(j) == 1L && 
+               length(i) == 1L)) {
+            stop("When you need to retrieve single values using drop, use single indexes")
```
Melkiades, line 1027: Before, in cases where subsetting was not only one element, it just did not care about drop=TRUE in the result. I changed that behavior so it is clearer how drop works

Melkiades: ref. issue #475

R/tt_compare_tables.R
```diff
@@ -51,7 +51,7 @@ trim_rows <- function(tt, criteria = all_zero_or_na) {
     rows <- collect_leaves(tt, TRUE, TRUE)
     torm <- vapply(rows, criteria,
                   NA, USE.NAMES = FALSE)
-    tt[!torm, , keep_topleft = TRUE]
+    tt[!torm, , keep_topleft = TRUE, keep_titles = TRUE, keep_fnotes = TRUE]
```
gmbecker, line 54: reindex references as well please

R/tt_pos_and_access.R
```diff
@@ -327,52 +327,90 @@ setMethod("tt_at_path<-", c(tt = "VTableTree", value = "TableRow"),
...
+#' \code{TRUE}.}
 #' }
 #' @param value Replacement value (list, `TableRow`, or `TableTree`)
+#' 
 #' @return a \code{TableTree} (or \code{ElementaryTable}) object, unless a
 #'   single cell was selected with \code{drop=TRUE}, in which case the (possibly
 #'   multi-valued) fully stripped raw value of the selected cell.
-#' @exportMethod [<-
+#' 
+#' @details the standard subsetting drops the information about title, subtitle,
```
gmbecker, line 346: Please use "by default" instead of "the standard" (here and generally) as its more concise and accurate in english

R/tt_pos_and_access.R
```diff
@@ -327,52 +327,90 @@ setMethod("tt_at_path<-", c(tt = "VTableTree", value = "TableRow"),
...
+#' main footer, provenance footer, and `topleft`. If only a column is selected and
+#' all rows are kept, the `topleft` information remains as default. Any referential
+#' footnote is kept whenever the subset table contains the referenced element. 
+#' Please note that subsetting always preserve the original order, even if provided
```
gmbecker, line 350: This should be an explicit note (a la `@note`) rather than simply in the text

R/tt_pos_and_access.R
```diff
@@ -327,52 +327,90 @@ setMethod("tt_at_path<-", c(tt = "VTableTree", value = "TableRow"),
...
+#' tbl[1, 2, keep_fnotes = TRUE]
+#' tbl[1, 2, keep_topleft = TRUE]
+#' 
+#' # Keeps the referential footnotes when subset contains them
+#' fnotes_at_path(tbl, rowpath = c("SEX", "M", "AGE", "Mean")) <- "important"
+#' tbl[4,1]
+#'
+#' # Note that order can not be changed with subsetting
+#' tbl[c(4,3,1), c(3,1)] # It preserves order and wanted selection
+#'
```
gmbecker, line 404: Add example using `reindex_refs= FALSE` and the default (`TRUE`) give different answers

R/tt_pos_and_access.R
```diff
@@ -493,16 +564,19 @@ setGeneric("subset_cols",
                     newcinfo = NULL,
                     keep_topleft = TRUE,
                     keep_titles = TRUE,
+                    keep_fnotes = TRUE,
```
gmbecker, line 567: I don't really see the need for this, if referential footnotes are on elements that remain after the subsetting, they should always be kept (but may need to be reindexed).

Can you tell me more about why you felt this should be added?

Unless this means _footer materials_, in which case I'm still not convinced why we would need a separate argument beyond titles, but I'm open to being convinced (if so the name will need to change though, fnotes in  a function/argument name means _referential footnotes_ in `rtables`)

R/tt_pos_and_access.R
```diff
@@ -939,42 +1003,49 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
           function(x, i, j, ..., drop = FALSE) {
     ## have to do it this way because we can't add an argument since we don't
     ## own the generic declaration
-    keep_topleft <- list(...)[["keep_topleft"]] ## returns NULL if not presesnt
```
gmbecker, line 942: no reason to remove this comment, please put it back

R/tt_pos_and_access.R
```diff
@@ -748,9 +809,11 @@ setMethod("subset_cols", c("LayoutColTree", "numeric"),
 ## label rows ARE included in the count
 subset_by_rownum <- function(tt,
                              i,
-                             keep_topleft = NA,
+                             keep_topleft = FALSE,
                              keep_titles = TRUE,
+                             keep_fnotes = NULL, 
                              ...) {
+    if (is.null(keep_fnotes)) keep_fnotes <- keep_titles
```
gmbecker, line 816: just set the default to `keep_titles`

R/tt_pos_and_access.R
```diff
@@ -939,42 +1003,49 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
           function(x, i, j, ..., drop = FALSE) {
     ## have to do it this way because we can't add an argument since we don't
     ## own the generic declaration
-    keep_topleft <- list(...)[["keep_topleft"]] ## returns NULL if not presesnt
+    keep_topleft <- list(...)[["keep_topleft"]]
```
gmbecker, line 1006: no reason not to do the `%||%` thing we do for all the other ones here... (and above)

tests/testthat/test-printing.R
```diff
@@ -298,16 +298,18 @@ test_that("Inset works for table, ref_footnotes, and main footer", {
...
 
   # Adding references
   # row_paths(tt)
   # row_paths_summary(tt)
+  # col_paths(tt)
+  # col_paths_summary(tt)
   txt1 <- "Not the best but very long one, probably longer than possible."
   txt2 <- "Why trimming does not take it out?"
   fnotes_at_path(tt, rowpath = c("SEX", "F", "AGE", "Mean")) <- txt1
-  fnotes_at_path(tt, rowpath = c("SEX", "UNDIFFERENTIATED")) <- txt2
```
gmbecker, line 309: This should work as is why was it removed? If it doesn't work thats a bug

tests/testthat/test-printing.R
```diff
@@ -298,16 +298,18 @@ test_that("Inset works for table, ref_footnotes, and main footer", {
 
   # Building the table and trimming NAs
   tt <- build_table(lyt, DM)
-  tt <- trim_rows(tt)
+  tt <- prune_table(tt)
```
gmbecker, line 301: adding prunning is good, but keep the trim_rows as well.

If they result in identical tables in this case, make that an additional test.

tests/testthat/test-subset-access.R
```diff
@@ -206,6 +220,21 @@ test_that("top_left retention behavior is correct across all scenarios", {
     expect_identical(top_left(tbl[1:2, 1:2]), character())
     expect_identical(top_left(tbl[1:2, 1:2, keep_topleft = FALSE]), character())
     expect_identical(top_left(tbl[1:2, 1:2, keep_topleft = TRUE]), tlval)
+    
+    # error for multiple elemented selected for dropping
+    expect_error(tbl[1:2, 1, drop = TRUE])
```
gmbecker, line 225: Not error, see above comment

tests/testthat/test-subset-access.R
```diff
@@ -187,15 +187,29 @@ test_that("Duplicate colvars path correctly", {
                      res)
 })
 
-test_that("top_left retention behavior is correct across all scenarios", {
+test_that("top_left, title, footers retention behaviors are correct across all scenarios", {
```
gmbecker, line 190: add reindex_refs here too so that all the basic checks for `[` behavior are in one place

gmbecker (changes requested):
Great job, please revise based the individual comments

R/tt_pos_and_access.R
Melkiades: I intended general footers (main and prov). I thought it was giving another option to the user in case they need to keep only one of the two. Even if it is not yet used I think it is natural to ask if you can keep only titles or only footers when subsetting. I will change this into `keep_footers` for now. Let me know what you think ^^

R/tt_pos_and_access.R
Melkiades: The problem here is that if we want to add the option to keep that information whenever `keep_topleft` is not set and all rows are selected, we need to keep the knowledge of not having it set until the first check (i.e. if all rows are selected). I am not against keeping it as a fixed default. In that case, though, I think it should be `TRUE`.

tests/testthat/test-printing.R
Melkiades: So here, even with the current fix the two calls, `trim_rows(tt)` and `prune_table(tt)`, do not give the same results. The difference is that the empty nesting levels are kept with `trim_rows` while pruning takes them out. For example from #476 the result keeps UNDIFFERENTIATED in one case while the other takes it out:
```
                     all obs
                     (N=356)
  ——————————————————————————
  F                         
    Mean              34.13 
  M                         
    Mean              34.32 
  U                         
  UNDIFFERENTIATED      
```

About this, I heard @BFalquet saying that sometimes `trim_rows` is preferred because it keeps more structure that then can be more easily cbinded. Let me know if you want me to file an issue for having these be the same or if I should only test these two different behaviors.

tests/testthat/test-subset-access.R
Melkiades: I do not see the comment. For changing this behavior, I created an issue in https://github.com/Roche/rtables/issues/475

tests/testthat/test-printing.R
Melkiades: I will add additional tests to test it as it at the moment but prune and trim produce different tables

R/tt_pos_and_access.R
gmbecker: The previous behavior was correct; this should not be an error. `drop =  TRUE` is ignored for `data.frame` when "dropping" doesn't make any sense (i.e., the multi-multi case):

```
> mtcars[1,1, drop = TRUE]
[1] 21

> mtcars[1,1, drop = FALSE]
          mpg
Mazda RX4  21

> mtcars[,1, drop = TRUE]
 [1] 21.0 21.0 22.8 21.4 18.7 18.1 14.3 24.4 22.8 19.2 17.8 16.4 17.3 15.2 10.4
[16] 10.4 14.7 32.4 30.4 33.9 21.5 15.5 15.2 13.3 19.2 27.3 26.0 30.4 15.8 19.7
[31] 15.0 21.4

> mtcars[,1, drop = FALSE]
                     mpg
Mazda RX4           21.0
Mazda RX4 Wag       21.0
Datsun 710          22.8
Hornet 4 Drive      21.4
Hornet Sportabout   18.7
Valiant             18.1
Duster 360          14.3
Merc 240D           24.4
Merc 230            22.8
Merc 280            19.2
Merc 280C           17.8
Merc 450SE          16.4
Merc 450SL          17.3
Merc 450SLC         15.2
Cadillac Fleetwood  10.4
Lincoln Continental 10.4
Chrysler Imperial   14.7
Fiat 128            32.4
Honda Civic         30.4
Toyota Corolla      33.9
Toyota Corona       21.5
Dodge Challenger    15.5
AMC Javelin         15.2
Camaro Z28          13.3
Pontiac Firebird    19.2
Fiat X1-9           27.3
Porsche 914-2       26.0
Lotus Europa        30.4
Ford Pantera L      15.8
Ferrari Dino        19.7
Maserati Bora       15.0
Volvo 142E          21.4

> mtcars[1, , drop = TRUE]
$mpg
[1] 21

$cyl
[1] 6

$disp
[1] 160

$hp
[1] 110

$drat
[1] 3.9

$wt
[1] 2.62

$qsec
[1] 16.46

$vs
[1] 0

$am
[1] 1

$gear
[1] 4

$carb
[1] 4

> mtcars[1:2, 1:2, drop = TRUE]
              mpg cyl
Mazda RX4      21   6
Mazda RX4 Wag  21   6
> 
```

`rtables` should do the same thing, not throw an error

gmbecker (changes requested):

R/tt_pos_and_access.R
```diff
@@ -939,42 +1024,46 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
-            x <- row_values(rw)[[1]]
-    }
-    if(!drop) {
+                         keep_titles = keep_titles,
+                         keep_footers = keep_footers)
+
+    # Dropping everything
+    if (drop && any(length(j) == 1L,
+                    length(i) == 1L)) {
+        rw <- collect_leaves(x, TRUE, TRUE)
```
gmbecker, line 1060: this looks like a behavior change, it no longer has the `[[1]]`, is that intentional?

gmbecker (changes requested):

R/tt_pos_and_access.R
Melkiades: Yes, it is as we discussed before. In the previous comment, you wanted to have `drop` behave as it works for `data.frame`. In the following, the two behaviors are compared.
```r
tbl <- basic_table() %>%
  split_cols_by("Species") %>%
  analyze("Sepal.Length", function(x) {
    list("m" = mean(x), "sd" = sd(x))
  },
  format = "xx.x"
  ) %>%
  build_table(iris) %>%
  print()
```
produces the following:
```
     setosa   versicolor   virginica
————————————————————————————————————
m     5.0        5.9          6.6   
sd    0.4        0.5          0.6  
```

Classical `data.frame` and `rtables` produces:
```
> df <- as.data.frame(matrix_form(tbl)$string[-1,][,-1], row.names = c("m", "sd"))
> df
    V1  V2  V3
m  5.0 5.9 6.6
sd 0.4 0.5 0.6
> df[, 1, drop = FALSE]
    V1
m  5.0
sd 0.4
> tbl[, 1, drop = FALSE]
     setosa
———————————
m     5.0  
sd    0.4 

> df[, 1, drop = TRUE]
[1] "5.0" "0.4"
> tbl[, 1, drop = TRUE]
[1] 5.0060000 0.3524897
```
I modified this behavior (i.e. having `drop` working also with vectors) after our discussions. You suggested in your comment to use the same behavior that data.frame-drop uses. I added the error in the first place because I got confused by the previous printing of the table in cases where an entire or partial column or row was in input for the drop. I believe it is better to adopt the same behavior of default data.frame-drop here and allow vectors to be dropped

gmbecker: I'm not sure we can do the vector for a column thing because we have to think about what happens in the case of label rows. WE'd either get a list with NULL entries in it (likely not particularly useful for the end user) or we'd get something where `NROW(tbl[,1, drop = TRUE]) != NROW(tbl)` which definitely seems like something we don't want.

Melkiades: Let me test these specific cases and get back to you. I did not think about that. Intuitively I would give back only numeric values in general as drop drops everything but in cases where there are label rows I would give back the UNdropped table with a warning. Do you think it is a reasonable behavior? What about when we have mixed output numeric, characters? Again, intuitively I would give back a mixed vector or list with "" for the label rows, so we are aligned with the tabulation principles that rtables has. In the next comment, I will show you both these scenarios and some possible solutions for these specific cases

gmbecker: I don't think we should do anything that doesnt' follow the `length(tbl[,i, drop = TRUE]) == NROW(tbl)` criteria. Its too likely to cause bugs or otherwise be a gotcha to be useful, in my opinion.

In practice that means very often ("almost always") giving back a list vector in the "extract a column" case, which I don't think will ultimately be very useful to people, because most non-trivial tables will have either label rows or content rows somewhere in their makeup, and content rows usually have multi-valued cells.

Melkiades: you are very right. I did not think well about that. Now it does return a vector only if the one selected cell has multiple values or the row/column does not have any label rows or multiple values in it. If the last two cases are tried, a specific warning for both is produced and the dropping does not happen (i.e. it keeps the table structure). What do you think? All expected cases are tested in the new tests

R/tt_pos_and_access.R
```diff
@@ -939,42 +1025,62 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
     nr <- nrow(x)
     nc <- ncol(x)
     i <- .j_to_posj(i, nr)
     j <- .j_to_posj(j, nc)
 
     ##  if(!missing(i) && length(i) < nr) {
-    if(length(i) < nr) { ## already populated by .j_to_posj
-        keep_topleft <- isTRUE(keep_topleft)
+    if(length(i) < nr) {## already populated by .j_to_posj
+        if (is.null(keep_topleft)) keep_topleft <- FALSE
```
gmbecker, line 1041: why/how would `keep_topleft` ever be `NULL`? should that be `is.na`? If so that is equivalent to the isTRUE form above, which is what it was doing previously

R/tt_pos_and_access.R
```diff
@@ -939,42 +1025,62 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
-    keep_topleft <- list(...)[["keep_topleft"]] ## returns NULL if not presesnt
+    keep_topleft <- list(...)[["keep_topleft"]] ## returns NULL if not present
     keep_titles <- list(...)[["keep_titles"]] %||% FALSE
+    keep_footers <- list(...)[["keep_footers"]] %||% keep_titles
     reindex_refs <- list(...)[["reindex_refs"]] %||% TRUE
-    if(is.null(keep_topleft))
-        keep_topleft <- NA
-
+    
+    # browser()
```
gmbecker, line 1033: Remove this please

R/tt_pos_and_access.R
```diff
@@ -939,42 +1025,62 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
+        if (is.null(keep_topleft)) keep_topleft <- FALSE
         x <- subset_by_rownum(x, i,
                               keep_topleft = keep_topleft,
-                              keep_titles = keep_titles)
-    } else {
-        keep_topleft <- !identical(FALSE, keep_topleft)
+                              keep_titles = keep_titles,
+                              keep_footers = keep_footers)
+    } else if (is.null(keep_topleft)) {
+        keep_topleft <- TRUE
```
gmbecker, line 1047: Why is `NULL` value of `keep_topleft` now translating to `TRUE` when it translates to `FALSE` above??

R/tt_pos_and_access.R
```diff
@@ -939,42 +1025,62 @@ setMethod("[", c("VTableTree", "numeric", "numeric"),
...
     }
+    
+    # browser()
```
gmbecker, line 1050: remove

R/tt_pos_and_access.R
```diff
@@ -1157,29 +1263,18 @@ setMethod("value_at", "LabelRow",
...
-    if(keep_topleft)
-        top_left(res) <- top_left(x)
-    else
-        top_left(res) <- character()
-    if(keep_titles)
-        res <- .copy_titles(res, x)
-    else
-        ## no titles
-        res <- .copy_titles(res, rtable(" "))
+    res <- .copy_titles_footers_topleft(res, x,
```
gmbecker, line 1273: This (`.h_t_body`) is so similar to `.copy_titles_footers_topleft` that I'm wondering if we need both functions?

gmbecker (changes requested):

R/tt_pos_and_access.R
Melkiades: This is very true! I did not think I could merge them. Now done

R/tt_pos_and_access.R
Melkiades: I thought about this and I cannot find a better solution. The feature here to maintain is that whenever only columns are subsetted, the topleft information, if not specified, is maintained. The original implementation sent down the `NA` value and then it was treated differently from the other parameters in different functions (columns or row subsetting). I think it is better to have the `NULL` or `NA` option only on the first level and assume it is a boolean in all other cases. To do so I have to keep it as "missing" until it is known if only columns are subsetted.

gmbecker (approved):

Issue #483, opened 2022-12-09
spl_context is missing row representing the root state (ie before first split applied)


gmbecker:
Relevant to the reverse-demo usecase.

PR #484, opened 2022-12-09
Ensure spl_context has row for root state with full data. #483

gmbecker:
@Melkiades please create branches in `tern` and `citril`/`chevron` and run their tests against this change, as it is technically a 'breaking change'. I say technically because it subtly changes the behavior of `.spl_context` (which almost no one is using) to have a new row at the top for the root state (ie before the first row split).


Melkiades:
I will take care of tern and tlg. @BFalquet can you take care of chevron/citril?

Melkiades:
> I will take care of tern and tlg. @BFalquet can you take care of chevron/citril?

Any update @BFalquet?

BFalquet:
> Any update @BFalquet?

Seem to be ok for the chevron side. One new function will be impacted but the fix seems easy.
=> go ahead with merge

PR #521, opened 2023-02-06
Allow 2d count (%) formats in column counts

### LINKED ISSUE #516: Add new format for `add_colcounts(-, format = "N=xx (100%)")`
Hi there,

I like rtables a lot and am always amazed by how lean the code for complicated tables can be using your package. One thing, my organization is overly focused on, is column counts being displayed as `N=134 (100%)`. The display of 100%, to indicate the reference for all following percentages, can currently not be achieved with rtables.

Is this something you would consider addressing in a future update?

Thank a lot and best regards,
Alex

edelarua:
Closes #516 

@gmbecker should I also add the requested `"N=xx (xx%)"` format as a valid 2d format in `formatters`?

### DIFF
--- a/R/tt_toString.R
+++ b/R/tt_toString.R
@@ -539,6 +539,17 @@ get_formatted_fnotes <- function(tt) {
     if (disp_ccounts(cinfo)) {
         counts <- col_counts(cinfo)
         cformat <- colcount_format(cinfo)
+
+        # allow 2d column count formats (count (%) only)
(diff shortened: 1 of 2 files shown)


gmbecker (changes requested):
This looks good but please add some tests for the new functionality.

gmbecker (approved):
Looks good. We need to expose this to basic_table, but I will take care of that. Thanks

PR #524, opened 2023-02-07
Disable unit test performance reports

cicdguy:
Disable unit test performance reports.


gmbecker (approved):
@cicdguy this is fine, thanks. I think a larger issue is whether we should have/place stock in these things generally (beyond rtables specifically) which was more my point when I @'ed you.

Anyway, thanks

PR #531, opened 2023-02-14
change default pagebreak from \s\n to \f yclose #530

### LINKED ISSUE #530: replace default page_break from "\s\n" to "\f"
requested by Liming in https://github.com/insightsengineering/rlistings/pull/78

shajoezhu:
Fixes #530


R/tt_export.R
```diff
@@ -166,7 +166,7 @@ export_as_txt <- function(tt, file = NULL,
                           paginate = .need_pag(page_type, pg_width, pg_height, lpp, cpp),
                           cpp = NULL,
                           lpp = NULL,
-                          ..., page_break = "\\s\\n",
+                          ..., page_break = "\\f",
```
Melkiades, line 169: I must have lost the discussion but why should we use these page break indicators? I mean how this was chosen?

Melkiades (approved):
I am fine with all of this but I do not understand why this was necessary in the first place. Anyway, @gmbecker fyi

Melkiades:
> I am fine with all of this but I do not understand why this was necessary in the first place. Anyway, @gmbecker fyi

Response: using \f allows users to directly copy output in word, and make a new page

gmbecker:
> > I am fine with all of this but I do not understand why this was necessary in the first place. Anyway, @gmbecker fyi
> 
> Response: using \f allows users to directly copy output in word, and make a new page

I'm still not convinced of this needing to be the default, given that the use case is quite specific, and the argument is already available (and can be wrapped as the default in chevron).

I don't want to merge this before we have a further discussion about it

gmbecker:
@cicdguy I dont' intend to merge this PR currently, but why am I not authorized to merge into rtables main branch?

PR #538, opened 2023-02-21
Update invalid `ref_group` error message

### LINKED ISSUE #499: Improve the error message of "bline %in% vals" [example still work in progress]

Please briefly describe your problem and, when relevant, the output you expect.
Please also provide the output of `utils::sessionInfo()` or
`devtools::session_info()` at the end of your post.

If at all possible, please include a [minimal, reproducible
example](https://stackoverflow.com/questions/5963269/how-to-make-a-great-r-reproducible-example).
The `rtables` team will be much more likely to resolve your issue if they are
able to reproduce it themselves locally.

Please delete this preamble after you have read it.

---

your brief description of the problem

```r
library(rtables)

# your reproducible example here

```

edelarua:
Closes #499

### DIFF
--- a/R/tt_dotabulation.R
+++ b/R/tt_dotabulation.R
@@ -1251,9 +1251,8 @@ setMethod("set_def_child_ord", "VarLevWBaselineSplit",
         }
     }
     if (!bline %in% vals) {
-        stop(paste(
-            "The ref_group value specified for split_cols_by was not present in the levels of", 
-            spl_payload(lyt), "in the data."
+        stop(paste0(
+            '"', bline, '"', ' was not present in the levels of ', spl_payload(lyt), ' in the data.'
         ))
     }
     pos <- match(bline, vals)
--- a/R/tt_dotabulation.R
+++ b/R/tt_dotabulation.R
@@ -1252,7 +1252,7 @@ setMethod("set_def_child_ord", "VarLevWBaselineSplit",
     }
     if (!bline %in% vals) {
         stop(paste0(
-            '"', bline, '"', ' was not present in the levels of ', spl_payload(lyt), ' in the data.'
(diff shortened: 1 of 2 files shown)


R/tt_dotabulation.R
```diff
@@ -1250,8 +1250,12 @@ setMethod("set_def_child_ord", "VarLevWBaselineSplit",
             vals <- levels(relevel(droplevels(vals), bline))
         }
     }
-
-    stopifnot(bline %in% vals)
+    if (!bline %in% vals) {
+        stop(paste(
```
gmbecker, line 1254: Please include both the missing ref group (`bline`) and the variable (`spl_payload(lyt)`, as you already have) in the error message.

gmbecker (changes requested):

R/tt_dotabulation.R
edelarua: I've added `bline` to the error message.

R/tt_dotabulation.R
```diff
@@ -1250,8 +1250,11 @@ setMethod("set_def_child_ord", "VarLevWBaselineSplit",
             vals <- levels(relevel(droplevels(vals), bline))
         }
     }
-
-    stopifnot(bline %in% vals)
+    if (!bline %in% vals) {
+        stop(paste0(
+            '"', bline, '"', ' was not present in the levels of ', spl_payload(lyt), ' in the data.'
```
gmbecker, line 1255: We still want to say that the missing level was supposed to be the reference group so that the user knows where to look to fix the error. Please do something along the lines of 

"Reference group (<bline>) was not present in levels of <spl_payload(lyt)> in the data."

gmbecker (changes requested):

R/tt_dotabulation.R
edelarua: Error message updated accordingly.

gmbecker (approved):

PR #576, opened 2023-03-09
Add split_label to rows_by_multivar, extra_args to both multivar

gmbecker:


Melkiades (approved):
Good to go for me if you think it is complete. I would only add an entry to the news (like misc.) ;)

PR #584, opened 2023-03-14
Connect `custom_split_funs` documentation

### LINKED ISSUE #561: `custom_split_funs` is not mentioned in the vignette and viceversa
I think we need it less stand alone so either we reintegrate it into make_split_fun or we extend it and connect it to the related vignette

edelarua:
Closes #561

### DIFF
--- a/R/split_funs.R
+++ b/R/split_funs.R
@@ -79,6 +79,9 @@ setGeneric(".applysplit_ref_vals",
 #' functions and modify either the incoming data before they are called or
 #' their outputs.
 #'
+#' @seealso [make_split_fun()] for a simplified interface for creating custom split 
+#' functions, and [split_funcs] for a variety of pre-defined split functions.
+#'
 #' @examples
 #' # Example of a picky split function. The number of values in the column variable
 #' # var decrees if we are going to print also the column with all observation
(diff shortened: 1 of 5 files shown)


R/split_funs.R
```diff
@@ -79,6 +79,9 @@ setGeneric(".applysplit_ref_vals",
 #' functions and modify either the incoming data before they are called or
 #' their outputs.
 #'
+#' @seealso [make_split_fun()] for a simplified interface for creating custom split
```
gmbecker, line 82: please change "a simplified interface" to "the API", as in "... for the API for creating custom split functions"

vignettes/split_functions.Rmd
```diff
@@ -291,6 +291,10 @@ The pre-processing component is useful for things such as manipulating
 factor levels, e.g., to trim unobserved ones or to reorder levels
 based on observed counts, etc.
 
+For further details on the structure of custom split functions, and an 
+example of a custom split function that does not implement `make_split_fun()`,
```
gmbecker, line 295: "does not implement" -> "not implemented via `make_split_fun()`

But a larger issue is I'm not sure we need/want this. I don't know of a single function that can't be done in `make_split_fun`...

Instead I'd say "for a more detailed discussion of what custom split functions do", or similar. And honestly, we likely need to merge `?custom_split_funs` into `?make_split_fun` at least partially, but we'll do that as a separate issue.

gmbecker (changes requested):

gmbecker (approved):

PR #585, opened 2023-03-15
Start deprecation cycle for `trim_zero_rows`

### LINKED ISSUE #562: `trim_zero_rows` is disconnected from documentation
This function is not documented in a vignette and it is missing examples. I think it should be under `trim_rows`.

edelarua:
Closes #562


shajoezhu (commented):
hey @edelarua , can you add life cycle badges to this?

edelarua:
@shajoezhu `rtables` doesn't currently use the `lifecycle` package, so you should open another issue if you want to add that into the package as it will involve more than just this function.

shajoezhu (approved):
lgtm! Thanks @edelarua

PR #589, opened 2023-03-17
Fix bug in `sort_at_path`

### LINKED ISSUE #588: Fix bug occurring when `cell_values` is used after `sort_at_path`
When `cell_values` is used to access a cell value in a table after using `sort_at_path` with a path including `"*"`, the following error occurs. This bug is caused by the labels being dropped while sorting with `"*"` in the sort path.

```
Error in tt_at_path(tt, rowpath) :
  Path appears invalid for this tree at step XXX
``` 

For example:

```r
tbl <- basic_table() %>%
    split_cols_by("ARM") %>%
    split_rows_by("RACE") %>%
    summarize_row_groups() %>%
    analyze("STRATA1") %>%
    build_table(DM)

scorefun <- function(tt) sum(unlist(row_values(tt)))

tbl <- sort_at_path(tbl, c("RACE", "*", "STRATA1"), scorefun)

cell_values(tbl, c("RACE", "BLACK OR AFRICAN AMERICAN", "STRATA1", "C"), c("ARM", "A: Drug X"))
```

Produces:
```
Error in tt_at_path(tt, rowpath) : 
  Path appears invalid for this tree at step BLACK OR AFRICAN AMERICAN
```

edelarua:
Closes #588

### DIFF
--- a/R/tt_sort.R
+++ b/R/tt_sort.R
@@ -194,6 +194,7 @@ sort_at_path <- function(tt,
                              ## informative error reporting.
                              .prev_path = c(.prev_path, backpath, paste0("* (", oldnames[i], ")")))
                 })
+            names(newkids) <- oldnames
             newtab <- subtree
             tree_children(newtab) <- newkids
             if(length(backpath) > 0) {


gmbecker (approved):
Nice catch

gmbecker:
@edelarua in the future please be sure to add a regression test in cases like this. I will do so for this bug

PR #595, opened 2023-03-23
Add test for `spl_variable`

### LINKED ISSUE #582: 'Variable Associated With a Split' is not covered by any test

edelarua:
Closes #582


gmbecker (approved):
This is fine, but the more relevant change is that there are places in the codebase that currently use spl_payload that should probably be using spl_variable (as it is more semantically meaningful, which is why i created it even though its basically a conditional wrapper around spl_payload). Once those are changed, we would have the coverage already. 

This test seems fine though.

PR #602, opened 2023-04-06
Keep annotations when rbinding two tables

### LINKED ISSUE #427: title dissapears when rbinding two rtables
```
library(rtables)

adsl <- ex_adsl

t1 <- basic_table(title = "my table") %>%
  split_cols_by("ARM") %>%
  analyze("AGE") %>%
  build_table(df = adsl)

t2 <- basic_table(title = "my table") %>%
  split_cols_by("ARM") %>%
  analyze("SEX") %>%
  build_table(df = adsl)


rbind(t1, t2)

```
There is no title. 

image

The need for binding rtables is quite high when we have the typical AE summary table where we have to deal with different filtered analysis ADAE datasets to get the target analysis population for subjects with at least one TEAE, AESI, etc. The best approach is to create subtables and then rbind them. 

image

```
> packageVersion("rtables")
[1] ‘0.5.2’
```

edelarua:
Annotations will be kept if only present in the first object or identical in all objects.

Closes #427


R/tt_compatibility.R
```diff
@@ -374,6 +374,21 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
 
     lapply(x, function(xi) chk_compat_cinfos(x[[1]], xi)) ##col_info(xi)))
 
+    rbind_annot <- list(title = "", subtitles = character(), footer = character(), pf = character())
+    # If rbinding 2 tables, annotations are retained if only present in one table or shared by both
+    if (!any(sapply(x, obj_name) == "rbind_root") & all(sapply(x, \(x) is(x, "ElementaryTable")))) {
```
gmbecker, line 379: No reason they would need to be Elementary tables here.

Also, why not rbind_root, all that means is that the table came from a previous rbind, doesn't speak to annotations at all

Finally, please avoid `\(x)`. I know its valid in modern versions of R but `function(x)` isn't that much longer and is easier to read and consistent with the codebase overall.

NEWS.md
```diff
@@ -1,5 +1,6 @@
 ## rtables 0.6.0.1
  * bumped devel version
+ * Joining 2 tables via `rbind` will retain annotations if only present in one table or shared between both tables.
```
gmbecker, line 3: We shouldn't use the word "joining" here as it has a specific meaning in data science, which isn't what we're doing.

Also "annotations" isn't specific enough, we would want to say "title and footer information" or something similar.

gmbecker (changes requested):
So this has been discussed before. The tension here is between consistency (annotations are always dropped) vs inconsistent but often convenient (annotations are sometimes dropped, and sometimes not).

There are 4 (the last two being variants of the same theme) possible behaviors that we have come up with

1. Annotations are always dropped (the current behavior AFAIR)
2. Annotations are kept if the first table has them and all other tables are NULL.
3. Annotations from the first table are always kept and annotations from others are ignored
4. Annotations which agree are kept, but any which don't agree are stripped
5. Annotations which agree, OR which are present in only one table and blank/NULL in all others are kept, all others are stripped.

I'm not convinced that (4/5) is the correct behavior, and here's why: Firstly it introduces uncertainty ("we always have to reannotate" -> "we may need to reannotate and our code probably has to check"), which is bad. Secondly, in cases where some of the elements of anntatoin match and others dont, its not clear stripping some annotations and keeping others would ultimately be the right behavior. Imagine two tables, both of which have titles "STUDY AWESOME Coolest Table", one of which has subtitle "stratified by stuff", the other has "stratified by thing". Getting a table with title "STUDY AWESOME Coolest Table" but _**no subtitle at all**_ when you rbind them sounds like its going to be annoying in practice. 

Finally, and perhaps most importantly, this shouldn't really happen. If two different tables have the same non-null annotations, then the annotations are wrong on (at least) one of them, and shouldn't be there.  They only way they'd have those matching title/footer materials is if they were put there intentionally as a work around, but in that case, said work around can simply be done _**after**_ the rbind.

Honestly, I'm much more inclined to think (2) above is the correct behavior than (4/5), currently.

edelarua:
I agree that implementing 2 instead of 4/5 is more straightforward for the user and I have changed the PR to reflect this/the other comments left above.

NEWS.md
edelarua: Updated accordingly.

R/tt_compatibility.R
edelarua: Updated/removed accordingly.

R/tt_compatibility.R
```diff
@@ -374,6 +374,25 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
...
+    all_titles <- lapply(x, function(x) if (is.null(attr(x, "main_title"))) "" else main_title(x))
+    if (all_titles[[1]] != "" & all(all_titles[[-1]] == "")) rbind_annot[["title"]] <- all_titles[[1]]
+
+    all_sts <- lapply(x, function(x) if (is.null(attr(x, "subtitles"))) character() else subtitles(x))
+    if (length(all_sts[[1]]) > 0 & all(sapply(all_sts[[-1]], function(x) length(x) == 0))) {
+      rbind_annot[["subtitles"]] <- all_sts[[1]]
+    }
+
+    all_footers <- lapply(x, function(x) if (is.null(attr(x, "main_footer"))) character() else main_footer(x))
+    if (length(all_footers[[1]]) > 0 & all(sapply(all_footers[[-1]], function(x) length(x) == 0))) {
```
gmbecker, line 388: Overall, I think retaining annotations should be an all or nothing thing. If any of the annotations disagree between two fo the tables, they no annotations should be perserved. If they ALL agree, or if all but the first table have no annotations at all, then the first table's annotations should be retained.

R/tt_compatibility.R
```diff
@@ -407,6 +433,11 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
 #' @param deparse.level numeric(1). Currently Ignored.
 #' @param \dots ANY. Elements to be stacked.
 #'
+#' @note 
+#' When objects are rbinded, titles and footer information are retained from the first object (if any exist)
```
gmbecker, line 437: Looks like this text is a mix of two different versions, please fix to accurately describe the behavior

gmbecker (changes requested):

R/tt_compatibility.R
```diff
@@ -374,6 +374,25 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
 
     lapply(x, function(xi) chk_compat_cinfos(x[[1]], xi)) ##col_info(xi)))
 
+    rbind_annot <- list(title = "", subtitles = character(), footer = character(), pf = character())
+    # Titles/footer info are (independently) retained from first object if all other objects have none
+    all_titles <- lapply(x, function(x) if (is.null(attr(x, "main_title"))) "" else main_title(x))
```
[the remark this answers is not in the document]
edelarua: If `main_title()` is used instead of `attr()` then an error occurs when binding with non-tables (`rrow()` objects, etc.) which don't have these attributes.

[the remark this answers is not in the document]
gmbecker: The fact that this happens to work for rtables objects is because slots are stored in an S4 object's attributes. We dont' want to use attr here.

There are a couple of to get around this that doesn't violate the abstraction.

The easiest is probably to do 

```
setMethod("main_title", "TableRow", function(x) "")
```

(and the same for the other title/footer getters). Then we don't even need the if statement.

Melkiades:
@gmbecker any news about this?

NEWS.md
```diff
@@ -8,7 +9,7 @@
  * Added checks to catch bad labels (with {}) and throw informative error.
  * Added `qtable` function to create a table with a single top-level structure in both row and column dimensions  involving faceting by 0 or more variables in each.
  * Added `as_result_df` function to flattern a table into a dataframe.
- * Added `sanitize_table_struct`, `validate_table_struct`, `find_degen_struct` to support degenerative table rendering.
+ * Added `sanitize_table_struct`, `validate_table_struct`, `find_degen_struct` to support degenerative table renderin
```
Melkiades, line 12:
```suggestion
 * Added `sanitize_table_struct`, `validate_table_struct`, `find_degen_struct` to support degenerative table rendering.
```

R/tt_compatibility.R
```diff
@@ -374,6 +374,30 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
...
+      !isFALSE(Reduce(function(x, y) if (identical(x, y)) x else FALSE, all_sts)))) {
+      rbind_annot[["subtitles"]] <- all_sts[[1]]
+    }
+
+    all_footers <- lapply(x, main_footer)
+    if (length(all_footers[[1]]) > 0 && (all(sapply(all_footers[[-1]], length) == 0) ||
+      !isFALSE(Reduce(function(x, y) if (identical(x, y)) x else FALSE, all_footers)))) {
+      rbind_annot[["footer"]] <- all_footers[[1]]
+    }
+
```
Melkiades, line 395: is there a way to improve readability here? also for other booleans

gmbecker: Yes, if we read this carefully its doing two things, we should factor each into a function, even if its only used once, for maintainability. I'll leave what those functions would be as practice of what we talked about today. @Melkiades @ayogasekaram @edelarua

Melkiades: I think now it is very clear! Thanks Emily

R/tree_accessors.R
```diff
@@ -2213,6 +2226,19 @@ setMethod("subtitles<-", "VTitleFooter",
...
+##' @export
+setMethod("subtitles", "TableRow",
+          function(obj) character())
+
+
+##' @rdname formatters_methods
+##' @export
+setMethod("subtitles<-", "TableRow",
+          function(obj, value) {
+              obj@subtitles <- value
```
gmbecker, line 2239: `TableRow` class does not have this slot. This function will break if it is ever called.

Same for all the other title/footer setters on `TableRow`

This also means our test coverage is insufficient

R/tt_compatibility.R
```diff
@@ -374,6 +391,19 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
 
     lapply(x, function(xi) chk_compat_cinfos(x[[1]], xi)) ##col_info(xi)))
 
+    rbind_annot <- list(title = "", subtitles = character(), footer = character(), pf = character())
```
gmbecker, line 394: Please use the slot names here as names of the list (ie prov_footer instead of pf) for ease of understanding.

Also, we need to talk about whether these checks should be independent or collective. My instinct is collective, but lets discuss it at our next standup. @anajens @edelarua @ayogasekaram @Melkiades please think on this and come with your thoughts collected.

R/tt_compatibility.R
```diff
@@ -374,6 +391,19 @@ rbindl_rtables <- function(x, gap = 0, check_headers = TRUE) {
...
+    # Titles/footer info are (independently) retained from first object if identical or missing in all other objects
+    all_titles <- sapply(x, main_title)
+    if (all_annots_identical(all_titles) || only_first_annot(all_titles)) rbind_annot[["title"]] <- all_titles[[1]]
```
gmbecker, line 397: This if statement (and those below it) aren't short enough to justify the single line form, please move the consequent (the thing that happens if the condition is true) to the next line here and below.

gmbecker (changes requested):

R/tt_compatibility.R
Melkiades: I think as @edelarua thought it respects the original user request and makes sense to keep only the first annotation or all equal independently anyway. Doing this collectively adds things to explain we do not need to. In other words, as you said in other context, this rbind checks would need to be smart and we do not want this too much. Even if we develop an intuitive strategy to do this collectively (which would add checks and complexity), it will drop information more than it does now

Melkiades:
@edelarua It seems that the error is related to the main_title getter for `LabelRow`. Is it for this reason you introduced the getters and setters (that I took out, sorry)? We should see to this ^^

Melkiades:
@gmbecker @shajoezhu, I think this can also go in

PR #614, opened 2023-04-21
Split Machinery and Tabulation - Dev Guide

### LINKED ISSUE #568: Outline Dev Guide Section - Split machinery
See `R/split_funs.R`

Melkiades:
Fixes #568 

fyi @ayogasekaram. We can meet anytime to exchange opinions and ideas on structure and content


inst/dev-guide/dg_tabulation.Rmd
```diff
@@ -0,0 +1,176 @@
...
+  rmarkdown::render(inputFile, encoding = encoding, output_dir = ".")})
+---
+
+```{r setup, include=FALSE}
+knitr::opts_chunk$set(echo = TRUE)
+```
+
+## Disclaimer
+
+Any code or prose which appears in a version of this vignette on the `main` branch of the repository may reflect a specific state of things that can be more or less recent. This guide reflects very important pieces of the split machinery that are unlikely to change. We anyway invite the reader to think always that the current code may have drifted from the following document, and it is always the best practice to read directly the code on `main`. (xxx we should insert it automatically)
```
Melkiades, line 23: I think we should insert this automatically. @edelarua @ayogasekaram you remember how best to add this from a common source?

edelarua: I think the way we did this previously (in `tern`?) only worked for Roxygen documentation, i.e. creating a doc with the blurb and reusing it with a few words changed in multiple places.

edelarua (approved):
Hi @Melkiades,

The dev guide sections look great! I made a few grammar/wording edits but nothing big and it was very easy to understand. I also applied the styler to the code blocks but didn't actually change any of the code. There's a few sections with "xxx" which I assume you will still want to edit, plus we will need to add some links in (probably better to do in one of the following PRs after publishing to the pkgdown site), but I think it's fine for now especially since we have the disclaimer.

Amazing work as always!! :)

inst/dev-guide/dg_split_machinery.Rmd
```diff
@@ -20,118 +20,127 @@ knitr::opts_chunk$set(echo = TRUE)
...
 basic_table() %>%
-    build_table(DM)
+  build_table(DM)
 ```
 
-In the following, we copied it so to allow the reader to go through the general structure with its enhanced comments and sections. Each section in the code reflects roughly a section of this vignette.
+In the following code, we copied the `do_split` function code to allow the reader to go through the general structure with enhanced comments and sections. Each section in the code reflects roughly one section of this article.
 
 ```{r, eval=FALSE}
-# rtables 6.0.2
```
Melkiades: ops... we should get to 1.0.0 at some time :D

inst/dev-guide/dg_split_machinery.Rmd
```diff
@@ -211,100 +224,106 @@ Formal class 'AllSplit' [package "rtables"] with 17 slots
...
-## default does nothing, add methods as they become
-## required
-setMethod("check_validsplit", "Split",
-          function(spl, df)
-    invisible(NULL))
+# rtables 0.6.2
+## Default does nothing, add methods as they become required
+setMethod(
+  "check_validsplit", "Split",
+  function(spl, df) invisible(NULL)
```
Melkiades, line 234: I think these were coming from the original source code

PR #615, opened 2023-04-21
Fix various issues for the str method for VTableTree

gmbecker:


gmbecker:
@Melkiades this fixes the issue we were having with str. It turns out we do have a str method, which was supposed to pass directlyd own but with a warning, but arguments were not being passed along.

I've also set the default level to 3, which is the right thing for VTableTree and actually makes it useful.  I've left a version of the warning in because even though it doesn't break R now, it still isn't the right thing for end users to do.

shajoezhu:
hi @gmbecker and @Melkiades , do we have an issue to track this change? Thanks

R/summary.R
```diff
@@ -439,16 +439,22 @@ is_empty_ElementaryTable <- function(x) {
...
 
 #' @export
-setGeneric("str", function(object)
+setGeneric("str", function(object, ...)
     standardGeneric("str"))
 
 #' @rdname int_methods
+#' @param max.level numeric(1). Passed to `utils::str`. Defaults to 3 for the
+#' VTableTree method, unlike the underlying default of `NA`. `NA` is *not*
+#' appropriate for `VTableTree` objects.
```
Melkiades, line 448: fantastic!!! I think this is very useful for devs

Melkiades (approved):
Good to go! Thanks, Gabe for taking care of this ;)

shajoezhu:
hi @gmbecker , thanks for making the changes. shall we merge this in already. Thanks

PR #617, opened 2023-04-28
Add vignette for cox regression analysis function

### LINKED ISSUE #616: Add vignette for cox regression analysis function

edelarua:
Closes #616


edelarua:
Hi @gmbecker,
Assuming that we don't want to add `tern` to Suggests, what would be the best way to use the `h_coxreg_extract_interaction` function here? i.e. Copy/paste the function code, create some simplified equivalent function, etc. 
I figure that detailing this step is not particularly relevant to this purpose of this vignette.

Melkiades:
> Hi @gmbecker, Assuming that we don't want to add `tern` to Suggests, what would be the best way to use the `h_coxreg_extract_interaction` function here? i.e. Copy/paste the function code, create some simplified equivalent function, etc.

To me, the best would be to use the real thing anyway. Do we have a helper function that checks and loads external packages? would this be a viable option?

vignettes/example_analysis_coxreg.Rmd
```diff
@@ -0,0 +1,380 @@
...
+    ## Store model in the caching environment
+    cache_env[[cov]] <- model
+  }
+  model
+}
+```
+
+## Creating the Analysis Function: `a_cox_summary`
+
+With our data prepared and helper function created, we can proceed to construct our analysis function `a_cox_summary`, which will be used to populate all of the rows in our table. In order to be used to generate both data rows (for interaction effects) and content rows (for main effects), we must create a function that can be used as both `afun` in `analyze` and `cfun` in `summarize_row_groups`. Therefore, our function must accept either `x` or `df` as its first parameter and also accept the `labelstr` parameter.
```
gmbecker, line 251: probably trim out the first part of this last sentence (all afuns and cfuns have to accept either x or df), the relevant bit is just the second half, that we make it accept labelstr so that it can be used in both contexts.

vignettes/example_analysis_coxreg.Rmd
```diff
@@ -0,0 +1,380 @@
...
+
+The arguments of our analysis function will be as follows:
+
+-   `df` - a `data.frame` of the full dataset required to fit the Cox regression model.
+-   `labelstr` - the `string` label for the variable being analyzed in the current row/column split context.
+-   `.spl_context` - a `data.frame` containing the `value` column which is used by this analysis function to determine the name of the variable/covariate in the current split. For more details on the information stored by `.spl_context` see `?analyze`.
+-   `stat` and `format` - `string`s that indicate which statistic column we are currently in and what format should be applied to print the statistic.
+-   `cache_env` - an `environment` object that can be used to store cached models so that we can prevent repeatedly fitting the same model. Instead, each model will be generated once per covariate and then reused.
```
gmbecker, line 259: say explicitly that this will be passed directly down to the `cached_model` helper we defined

vignettes/example_analysis_coxreg.Rmd
```diff
@@ -0,0 +1,380 @@
...
+-   `cov_main` - a `logical` value indicating whether or not the current row is summarizing covariate main effects.
+
+The analysis function works within a given row/column split context by using the current covariate (`cov`) and the `cached_model` function to obtain the desired Cox regression model. From this model, the `h_coxreg_extract_interaction` function is able to extract information/statistics relevant to the analysis and store it in a `data.frame`. The rows in this `data.frame` that are of interest in the current row/column split context are then extracted and the statistic to be printed in the current column is retrieved from these rows. Finally, the formatted rows containing the value of this statistic are returned with appropriate labelling. For more detail see the commented function code below, where the purpose of each line within `a_cox_summary` is described.
```
gmbecker, line 262: We don't return rows here, we return whats called a "VerticalRowsSection"  object (thats what in_rows acts as a constructor for), so please change the description of what is returned. 

VerticalRowsSection objects are basically stacks of _cells_, rather than full rows, they need to be placed side by side and then sliced horizontally to construct rows, but users defining analysis functions don't really need to know that.

gmbecker (changes requested):
Looks good just a couple of tweaks and we can get this in for the upcoming CRAN release

gmbecker (approved):
I think some wording changes are warranted, but I will try to make those as a separate PR and we can go through them as a way of working on technical writing.

PR #626, opened 2023-05-08
Update token name

walkowif:
Add changes from https://github.com/insightsengineering/idr-tasks/issues/543.


shajoezhu (approved):
Thanks @walkowif , approve to unblock the PR

PR #631, opened 2023-05-12
Fixes for exporter pagination migration

### LINKED ISSUE #627: import export_as_txt from formatters, and remove the current rtables export_as_txt, use staged dependencis 96_exporter_pagination_migration

Melkiades:
Fixes #627 

@gmbecker we need to talk about brackets! I found maybe a solution to the `devtools` AND check warnings caused by the base `[` and `[<-` reexports (apparently roxygen creates automatically reexport aliases when you directly export another namespace function, thus creating two .rd with reexports aliases but in different files, hence warning in checks). I went on reading how `tibble` does the same and it seems that uses S3 generics instead but w/o any reexports of base functions. So I tried to just remove the reexports and every test passes fine as brackets are dependent on the S4 generics as you designed them. 

I still remember though that we already mentioned this, so we need to discuss this solution (weirdly it works btw, very unexpectedly).


NAMESPACE
```diff
@@ -2,8 +2,6 @@
 
 S3method(print,CellValue)
 S3method(print,RowsVerticalSection)
-export("[")
-export("[<-")
```
Melkiades, line 6: these where previously reexported from base r @gmbecker

R/zzz_constants.R
```diff
@@ -7,6 +7,8 @@
                   assign.env = ns)
     delayedAssign("EmptyElTable", ElementaryTable(),
                   assign.env = ns)
+    delayedAssign("empty_table", rtable(" "),
+                  assign.env = ns)
```
Melkiades, line 11: I can probably use this for testing empty tables

man/brackets.Rd
```diff
@@ -22,10 +21,6 @@
 \alias{[,VTableTree,character,character-method}
 \alias{[,VTableTree,missing,numeric-method}
 \alias{[,VTableTree,numeric,numeric-method}
-\alias{bracket_subsetting}
-\alias{[}
-\alias{reexports}
-\alias{[<-}
```
Melkiades, line 28: @gmbecker This was causing the warning

man/brackets.Rd
```diff
@@ -175,12 +169,3 @@ tbl[c(4, 3, 1), c(3, 1)] # It preserves order and wanted selection
...
 }
-\keyword{internal}
-\description{
-These objects are imported from other packages. Follow the links
-below to see their documentation.
-
-\describe{
-  \item{base}{\code{\link[base:Extract]{[}}, \code{\link[base:Extract]{[<-}}}
-}}
-
```
Melkiades, line 186: This is the automatic reexports documentation I believe. In practice, it was generated inside this file in a non-standard way.

.pre-commit-config.yaml
```diff
@@ -9,7 +9,7 @@ repos:
         additional_dependencies:
         -    methods
         -    magrittr
-        -    insightsengineering/formatters
+        -    insightsengineering/formatters@96_exporter_pagination_migration
```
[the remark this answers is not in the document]
gmbecker: We temporarily do to get precommit to work. we will need to remove it again after merge in both packages happens.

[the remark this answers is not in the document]
gmbecker: Or more accurately, after only merge in formatters happens.

NAMESPACE
shajoezhu: we don't need them anymore? @Melkiades @gmbecker

R/tt_pos_and_access.R
```diff
@@ -1352,17 +1352,3 @@ setMethod("tail", "VTableTree",
...
-#' @name bracket_subsetting
-#' @export
-#' @usage
-#' x[i, j, ...]
-base::`[`
-
-#' @rdname brackets
-#' @export
-#' @usage
-#' x[i, j, ...] <- value
```
gmbecker, line 1365: Have you visually checked the pkgdown since you made this change, @Melkiades? I vaguely remember that may have been why we did this. @shajoezhu

Melkiades: You are very right! Now, I remember that, and it is true, the methods are automatically printed with the whole signature, being less than readable. I think though to have found a solution. See comment below

Melkiades:
I tried many combinations of inputs and `roxygen2` calls. Then, I checked how they do in `tibble` [here in sub.R](https://github.com/tidyverse/tibble/blob/7ec41e4513f076e82cfcd207681cf99422413cb9/R/sub.R), and it seems that their way is to have only one function per overloading operator (they use S3 generics as base R) and I thought to do the same here. 

tibble:
Screenshot 2023-05-15 152245

Before rtables:
Screenshot 2023-05-15 152919

Now rtables:
Screenshot 2023-05-15 153000

Information about the input values can be dispatched through the documentation. Here is the old reference page on main:
Screenshot 2023-05-15 153158
Screenshot 2023-05-15 153220

Here is the new one from rstudio on this branch:
Screenshot 2023-05-15 152159

The main change in practice is that S4 methods are no more responsible for dispatching the call to different functions, depending on the input indices (e.g. if they are logical, character, or missing) but they are directly dispatched inside the function.

Let me know @gmbecker what you think about this solution. I personally think that makes also the reference page clearer and more compact.

NAMESPACE
Melkiades: These were base r subsetting reexported. It was not strictly necessary but it was a good solution to the older problem of the reference page

shajoezhu:
hi @gmbecker , can you take a look at this issue please https://github.com/insightsengineering/rtables/issues/634

DESCRIPTION
```diff
@@ -24,7 +24,7 @@ URL: https://github.com/insightsengineering/rtables,
     https://insightsengineering.github.io/rtables/
 BugReports: https://github.com/insightsengineering/rtables/issues
 Depends: 
-    formatters (>= 0.4.0),
+    formatters (>= 0.4.1.9002),
```
shajoezhu, line 27:
```suggestion
    formatters (>= 0.4.1.9003),
```

shajoezhu:
brilliant! Thanks @gmbecker ! this is good to go?

gmbecker:
its not ready to cut the tag quite yet but yeah, its ready to merge into main in the leadup to that @shajoezhu

gmbecker (approved):

Issue #636, opened 2023-05-16
Bump version to 0.6.1, update NEWS.md, DESCRIPTION, check the README


shajoezhu:
blocked by

- [x] https://github.com/insightsengineering/rtables/pull/631
- [x] staged.dependencies to formatters, e.g. branch name 636_rtable_vbump@149_tag_release@main

Issue #607, opened 2023-04-19
new release tag, for OCEAN, by 2023-05-19


shajoezhu:
blocked by 

- [x] https://github.com/insightsengineering/rtables/issues/606
- [x] https://github.com/insightsengineering/rtables/pull/631

# TODOs

- [x]  Upstream formatters version bump to 0.5.0
- [x] #636
- [x]  Tag the update as a release candidate, v0.6.1-rc1
- [x]  Submit v0.6.1-rc1 for validation
- [x]  Get v0.6.1-rc1 validated, retag as v0.6.1
- [x]  Update downstream package dependencies to (>=0.6.1) in chevron
- [x]  Update downstream package dependencies to (>=0.6.1) in citril

PR #641, opened 2023-05-18
Fix font_size default. add r2rtf to Suggests. vbump. NEWS

gmbecker:
[skip vbump]

### DIFF
--- a/R/tt_export.R
+++ b/R/tt_export.R
@@ -121,7 +121,7 @@ path_enriched_df <- function(tt, path_fun = collapse_path, value_fun = collapse_
 ### Migrated to formatters.
 
 #' @importFrom formatters export_as_txt
-#' 
+#'
 #' @examples
 #' lyt <- basic_table() %>%
 #'   split_cols_by("ARM") %>%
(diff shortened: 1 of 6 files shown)


shajoezhu:
Hi @gmbecker , I would suggest to go in after #636 and #607

gmbecker:
> Hi @gmbecker , I would suggest to go in after #636 and #607 , we can better plan this and downstream testing during the next PI

If this were a different change I would agree but we don't want to make breaking changes to default font_size twice in a row. We are already committed to unifying font_size as having a default of 8, so we really should put this in now and tear the bandaid off. @shajoezhu

R/tt_export.R
```diff
@@ -504,86 +504,3 @@ export_as_pdf <- function(tt,
     }
      list(file = file, npages = npages, exceeds_width = exceeds_width, exceeds_height = exceeds_height, lpp = lpp, cpp = cpp)
 }
-
-.margin_lines_to_in <- function(margins, font_size, font_family) {
```
shajoezhu, line 508: hi @gmbecker , this was part of the migration right?

shajoezhu (approved):
lgtm! Thanks @gmbecker

R/tt_export.R
gmbecker: Yes, its no longer needed because we are now interpreting margins as inches values directly.

PR #644, opened 2023-05-25
fix major regressions in page_by functionality with new pag machinery

gmbecker:
[skip vbump]


tests/testthat/test-pagination.R
```diff
@@ -29,6 +29,26 @@ test_that("Page by splitting works", {
             split_rows_by("ARM", page_by = TRUE) %>%
             analyze("AGE")},
         "page_by splits cannot be nested within non-page_by splits")
+
+    lyt2 <- basic_table(title = "main title",
+                        subtitles = "subtitle",
+                        main_footer = "main footer",
+                        prov_footer = "provenance footer") |>
+    split_cols_by("ARM") |>
```
Melkiades, line 37: `|>` should we be consistent by using the other pipe?

gmbecker: Honestly, eventually we should probably convert everything over to |> which will allow us to trade a hard magrittr dependency for depending on R >= 4.1.0. 

Probably reasonable to do in 1-2 more releases of R (so either this year or next year). For now I've changed it though.

Melkiades (approved):
Lgtm! Thanks Gabe! :)

PR #650, opened 2023-06-07
specify min deps

pawelru:
WIP


.github/workflows/integration-tests.yaml
```diff
@@ -5,6 +5,9 @@ on:
   schedule:
     - cron: '45 3 * * 0'
   workflow_dispatch:
+  push:
+    branches:
+      - min_deps
```
pawelru, line 10: need to revert it once completed
```suggestion
```

DESCRIPTION
```diff
@@ -23,29 +23,41 @@ License: Apache License 2.0 | file LICENSE
...
+    testthat (>= 3.0.4),
+    tibble (>= 3.1.0),
     tidyr,
-    xml2
-VignetteBuilder: 
+    xml2 (>= 1.1.0)
+VignetteBuilder:
     knitr
+Remotes:
+    insightsengineering/formatters@*release,
```
gmbecker, line 49: I don't see a reason to add this, we should talka bout it before merging this in.

Generally I'm against specifying dependencies should be installed from specific places, as it is fundamentally the wrong approach and the control of that belongs in the hands of the end user (and changing where deps come from should NOT involve editing the package itself, which it does in the ill-advised remotes paradigm)

gmbecker (changes requested):

DESCRIPTION
pawelru: changed this into `Config/Needs/verdepcheck` field of DESCRIPTION file

shajoezhu:
hi @pawelru , I was wondering if I could check the status for this PR? if this is still ongoing?

pawelru:
This is to be closed soon. @averissimo is re-doing this now from scratch and I keep it live so that he can take some inputs out of it.

PR #654, opened 2023-06-09
initialization of degenerate tables' support

### LINKED ISSUE #566: prune_table returns NULL if all rows pruned
prune_table, according to documentation, shouls return a TableTree object, but it could also return NULL.

I wonder if we should return the table with no contents but the header? (is it a leagal rtable object?)

```{r}
adsl <- ex_adsl
levels(adsl$SEX) <- c(levels(ex_adsl$SEX), "OTHER")

tbl_to_prune <- basic_table() %>%
  split_cols_by("ARM") %>% 
  split_rows_by("SEX") %>% 
  summarize_row_groups() %>%
  split_rows_by("STRATA1") %>%
  summarize_row_groups() %>%
  analyze("AGE") %>%
  build_table(adsl[-c(1:400),])

tbl_to_prune %>% prune_table
```

Melkiades:
Fixes #566


tests/testthat/test_empty_tables.R
```diff
@@ -0,0 +1,48 @@
...
+    summarize_row_groups() %>%
+    analyze("AGE") %>%
+    append_topleft(c("Some", "    thing"))
+
+tbl_to_prune <- lyt %>% 
+    build_table(adsl[-c(1:400),])
+
+fnotes_at_path(tbl_to_prune, rowpath = c("F")) <- "agoraphobia"
+fnotes_at_path(tbl_to_prune, rowpath = NULL,
+               colpath = c("ARM", "A: Drug X")) <- "talassophobia"
```
Melkiades, line 23: This causes following error:
```
> pruned_tbl <- tbl_to_prune %>% prune_table
> pruned_tbl
Error in h(simpleError(msg, call)) :
error in evaluating the argument 'x' in selecting a method for function 'toString': attempt to set an attribute on NULL
```

tests/testthat/test_empty_tables.R
```diff
@@ -0,0 +1,48 @@
...
+})
+
+test_that("degenerate table can keep information if subset", {
+    pruned_tbl <- tbl_to_prune %>% prune_table
+    subset_tbl <- tbl_to_prune[-c(1:nrow(tbl_to_prune)), ,
+                 keep_topleft = TRUE, 
+                 keep_titles = TRUE, 
+                 keep_footers = TRUE]
+    
+    expect_identical(pruned_tbl, subset_tbl)
```
Melkiades, line 47: it fails because subsetting removes titles and footers anyhow

gmbecker:
I'm really not sure we should be making it easier for users to get their hands on a degenerate table while we don't formally support degenrate tables. And, to reiterate what I always say when this comes up, we do not currently support degenerate tables. They give undefined behavior across the api. That is likely to change eventually but it has not changed yet.

Also, honestly, pruning a table and getting 0 rows back should be at *least* a warning, if not an error, in my opinion. That is going to be a bug much much more often than it is the "correct" outcome.

Melkiades:
Closed after `assert_table_struct` and sanitize

Issue #517, opened 2023-02-02
Add `alt_counts_df` info to `.spl_context` or analyze parameters


edelarua:
When using `add_combo_levels` and creating a new level combining 2 arms, users may need to access `.N_col` for each of the combined arms when creating custom analysis functions (e.g. when calculating proportion difference between arms). This feature would be useful in Risk Difference column calculations, which are used in several safety tables.

This is not easily accomplished when using `alt_counts_df`, as only the sum of the `.N_col`s is accessible. This could be resolved by providing individual `.N_cols` for each of the combined levels or by providing `alt_counts_df` in `.spl_context` or as analyze parameters, or by some other similar solution.

PR #665, opened 2023-06-26
adding col split names and paths clearly to `.spl_context`

### LINKED ISSUE #539: Better information about column position in .spl_context
Currently, we can only imperfectly determine which column we are in by comparing the associated subsets to the current subset. 

We should have a better idea of where we are in column space. This will make e.g., creating analysis functions which create difference columns, etc easier.

Should probably happen at the same time as #517 so that we don't have to mess with the split_context machinery twice since doing so will be a breaking change (though one that is unlikely to actually break anything currently)

Melkiades:
Fixes #539 and #517


vignettes/clinical_trials.Rmd
```diff
@@ -497,6 +499,7 @@ We start by defining an events summary function:
 
 ```{r}
 s_events_patients <- function(x, labelstr, .N_col) {
+    # browser()
```
Melkiades, line 502:
```suggestion
```

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,256 @@
+context("Analysis functions (afun)")
+
+test_that(".spl_context contains information about the column split", {
```
Melkiades, line 3: workaround for doing sub multivar split @BFalquet

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,256 @@
...
+    tol <- 0.02
+    DM_tmp_F <- DM_tmp %>% filter(SEX == "F", STRATA1 == "B")
+    expect_equal(tbl[4, 1, drop = TRUE], mean(DM_tmp_F$BMRKR1), tolerance = tol)
+    expect_equal(tbl[4, 3, drop = TRUE], sd(DM_tmp_F$BMRKR1), tolerance = tol) # xxx why -0.011?
+})
+
+test_that(".spl_context contains col information and multivars works", {
+    skip()
+    DM_tmp <- DM %>% # xxx file an issue for this 
+                     # xxx add documentation for not supported + error
```
Melkiades, line 43: buggy, to understand why does not work

Melkiades: This is actually related to the previous working but inelegant solution (test 1)

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,256 @@
...
+        split_cols_by(var = "method") %>%
+        split_cols_by("SEX", split_fun = drop_split_levels) %>%
+        analyze(vars = "BMRKR1", afun = analysis_fun_fin, format = "xx.xxx")
+    
+    expect_silent(tbl <- lyt %>% build_table(DM_tmp))
+    
+    tol <- 0.02
+    DM_tmp_F <- DM_tmp %>% filter(SEX == "F", STRATA1 == "B")
+    expect_equal(tbl[4, 1, drop = TRUE], mean(DM_tmp_F$BMRKR1), tolerance = tol)
+    expect_equal(tbl[4, 3, drop = TRUE], sd(DM_tmp_F$BMRKR1), tolerance = tol) # xxx why -0.011?
```
anajens, line 37:
```suggestion
    DM_B_F <- DM %>% filter(SEX == "F", STRATA1 == "B")
    expect_equal(tbl[4, 1, drop = TRUE], mean(DM_B_F$BMRKR1))
    expect_equal(tbl[4, 3, drop = TRUE], sd(DM_B_F$BMRKR1))
```

anajens: Either need to filter original DM before calculating stats or need to subset by `method`. Doesn't make a difference for mean() but it affects sd().

anajens:
When I print the `.spl_context` data frame I get an error:
image

I'm pretty sure it's because the data.frame is not liking the expression in `cur_col_expr`. What's the rationale for including it?

This is my quick initial feedback, I'm still thinking about the scope of changes. I'm off tomorrow but will do a deeper dive on Monday.

gmbecker:
> When I print the `.spl_context` data frame I get an error: image

I get the same. 

> 
> I'm pretty sure it's because the data.frame is not liking the expression in `cur_col_expr`. What's the rationale for including it?

It is because of the expression. That needs to be wrapped in a list, because technically an expression is an object that has length (>1), so R is unhappy because that length doesn't match the number of rows indicated by the other columns.

The reason for including that expression is that unlike row faceting, the subsetting for column faceting is 'virtual', ie modeled, by an expression, until the point when the analysis/content function is actually invoked. 

The logic is (roughly) the following:

1. Subset based on the first row split into subsets corresponding to k_i facets
2. Step "into" each facet, separately (looping over them)
  a.  check if there is a content function, if so create the content row(s) for our current facet 
  b. Check if there is more splitting to be done, if so go to 1 and do it again recursively, if not advance to step c
  c. create the analysis rows for our current facet

So for step 2c, the incoming information is the row-facet data subset (a data.frame) and a set of column information.

The columns are modeled a couple different ways (see the `InstantiatedColumnInfo` class), but the most important for our purposes here is that for each leaf column, a subsetting expression is cached in that object. 

So to create the rows for our table, we first create what are called RowsVerticalSections, which are vertical stacks of cells, essentially, and are what are returned by `in_rows`.

We do this by looping over _**the column subsetting expressions**_ and applying them to our current row-facet data subset to get the subset for our individual facet-pane (which collapses to individual cells if your analysis function does not create multiple rows).

So at the time the split context is being created (or possibly when the column information is being appended to it, I haven't looked at Davide's implementation of that in detail yet), we have the row-facet dataframe, and an expression which was evaluated within that row-facet dataframe to get the input (either df, or one of its columns as x) passed to the analysis function. 

Technically, we could give a logical vector (ie after the evaluation I mention above).  I'm somewhat wary of that because they are not general, they only apply to the specific subset used to create them, whereas the expressions can be applied to any subset of the overal incoming data.

Anyway, that was very long-winded (@Melkiades please capture this and incorporate it into your dev-guide section), but hopefully it was informative.
> 
> This is my quick initial feedback, I'm still thinking about the scope of changes. I'm off tomorrow but will do a deeper dive on Monday.

gmbecker:
Also @anajens I had not heard you were back to working on this stuff. Welcome back. We should probably meet, a fair bit has changed (though the foundations you helped us design are largely intact)

R/colby_constructors.R
```diff
@@ -871,11 +883,62 @@ split_rows_by_cutfun <- function(lyt, var,
...
+#'     defined}
+#'   \item{.ref_full}{data.frame or vector of subset corresponding to the
+#'     `ref_group` column without subsetting defined by row-splitting. Optional
+#'     and only required/meaningful if a `ref_group` column has been defined}
+#'   \item{.in_ref_col}{boolean indicates if calculation is done for cells
+#'     within the reference column}
+#'   \item{.spl_context}{data.frame, each row gives information about a
+#'     previous/'ancestor' split state. See \code{\link{spl_context}}}
+#'   \item{.alt_counts_df}{list of data.frames, i.e. the `alt_count_df` after 
+#'     row splitting. Each element of the list represent a column-wise split.}
```
anajens, line 925: Currently `.alt_counts_df` is a single data.frame and not a list. Is there a use case for this argument? The naming is confusing since `.alt_counts_df` gives the `alt_counts_df` after the row split but typically `alt_counts_df` is typically related to the column N's and column splits.

R/colby_constructors.R
```diff
@@ -871,11 +883,62 @@ split_rows_by_cutfun <- function(lyt, var,
...
+#'   \item{.all_col_exprs}{vector of expressions. Each of them represents a
```
anajens, line 926:
```suggestion
#'   \item{.all_col_exprs}{list of expressions. Each of them represents a 
```

tests/testthat/test-tab_afun_cfun.R
anajens: Please update this test since it's incomplete.

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,256 @@
...
+                alt_df2 <- .alt_counts_df %>% 
+                    filter(ARM == "C: Combination",
+                           COUNTRY == .spl_context$cur_col_split_val[[1]][2])
+                
+                # Super manual extraction
+                alt_df1b <- alt_counts_df %>% 
+                    filter(ARM == "A: Drug X",
+                           COUNTRY == .spl_context$cur_col_split_val[[1]][2],
+                           SEX == .spl_context$value[3])
+                alt_df2b <- .alt_counts_df %>%
```
anajens, line 114:
```suggestion
                alt_df2b <- alt_counts_df %>% 
```

gmbecker: This looks correct, and suggests a larger change for increased maintainability (good eye/nice debugging to @anajens at catching this as it is).

Please change the argument name to the `n_wrapper_alt_df` function constructor to something easily distinguishable from `.alt_counts_df` (the analysis function argument name), e.g. (`full_alt_df` or `altdf` or `a_counts_df`) Basically anything that doesn't differ by only the smallest printed glyph (.).

Melkiades: renamed to `.alt_df` as it is not full and does not contain counts

tests/testthat/test-tab_afun_cfun.R
Melkiades: Now I will remove this test. Errors are flagged in #689 #690 #691. I think the first test is still worth keeping though. What do you think?

R/colby_constructors.R
Melkiades: Yes, you are right. I will change the doc. For a bit of context (pun not intended), in the beginning, I thought to have `.alt_df` see all the other row splits and not only the current one but it is out of scope for what Emily is using it and still, if this is the case, it would go into `.spl_context`. 

So there are two levels of faceting/subsetting here that someone might want to access: 1. all the same level facets, 2. all the hierarchical splits facets (currently what `full_df` is doing in `.spl_context`). The expected use needs neither, but only the current row split data frame. This, in turn, needs to be subsetted by different column expressions (hence all_col_exprs) for getting the right statistics (it is necessary for a case in which there is a combo df)

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,233 @@
...
+context("Content functions (cfun)")
+
+test_that(".alt_df appears in cfun but not in afun.", {
+    # Adding STRATA2 col to DM for alt_counts_df col split
+    alt_tmp <- DM %>% left_join(ex_adsl %>% 
+                                    mutate(ID = paste0("S", seq_len(nrow(ex_adsl)))) %>% 
+                                    select(ID, STRATA2))
+    
+    afun_tmp <- function(x, ...) rcell(mean(x), label = "MeAn", format = "xx.x")
+    cfun_tmp <- function(x, .alt_df, labelstr, .N_col,
```
anajens, line 189: When I print the arguments like `.alt_df` I see a warning:

```
Warning messages:
1: In get(object, envir = currentEnv, inherits = TRUE) :
  restarting interrupted promise evaluation
```

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,233 @@
...
+    
+    expect_identical(nrow_manual, spl_ctx_cnt)
+})
+
+test_that("Error localization for missing split variable when done in alt_count_df", {
+    # Error we want to happen
+    afun_tmp <- function(x, .alt_df, ...) mean(x)
+    lyt_col <- basic_table() %>% split_cols_by("ARMCD") %>% analyze("BMRKR1", afun = afun_tmp)
+    lyt_row <- basic_table() %>% split_rows_by("ARMCD") %>% analyze("BMRKR1", afun = afun_tmp)
+    expect_error(lyt_col %>% build_table(ex_adsl, alt_counts_df = DM))
```
anajens, line 155: Error here is as below. I don't think it should mention `df` since the problem is `alt_counts_df`? 
Also is there a way to make it read nicer like "alt_counts_df:  variable\\(s\\) \\[ARMCD\\] not present in data"?
```
Error in FUN(X[[i]], ...) :
alt_counts_df (or df) appears incompatible with column-split structure. Offending column subset expression: (!is.na(ARMCD) & ARMCD %in% "ARM A")
Original error message: object 'ARMCD' not found
```

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -0,0 +1,233 @@
...
+    lyt <- basic_table(show_colcounts = TRUE) %>% 
+        split_cols_by("SEX", split_fun = keep_split_levels(c("F", "U"))) %>% 
+        split_cols_by("STRATA2", split_fun = keep_split_levels("S1")) %>% 
+        split_rows_by("STRATA1", split_fun = keep_split_levels("A")) %>% 
+        summarize_row_groups(var = "BMRKR1", cfun = cfun_tmp) %>%
+        split_rows_by("ARMCD") %>% 
+        analyze("BMRKR1", afun = afun_tmp)
+    
+    expect_silent(lyt %>% build_table(ex_adsl))
+    expect_error(lyt %>% build_table(ex_adsl, alt_counts_df = DM))
```
anajens, line 230: Similar to above, would be nice to improve the error.
```
Error in FUN(X[[i]], ...) : 
  alt_counts_df (or df) appears incompatible with column-split structure. Offending column subset expression: ((!is.na(SEX) & SEX %in% "F")) & (!is.na(STRATA2) & STRATA2 %in% 
Original error message: object 'STRATA2' not foundalt_counts_df (or df) appears incompatible with column-split structure. Offending column subset expression:     "S1")
Original error message: object 'STRATA2' not found
```

R/tt_dotabulation.R
```diff
@@ -1008,7 +1104,9 @@ build_table <- function(lyt, df,
     lyt <- set_def_child_ord(lyt, df)
     lyt <- fix_analyze_vis(lyt)
     df <- fix_split_vars(lyt, df, char_ok = is.null(col_counts))
-
+    # This check should be added to cinfo and sent everywhere but for the moment
+    # there is little advantage in this as there are no other cases of this checks.
```
Melkiades, line 1108:
```suggestion
    # there is little advantage in this as there are no other cases for these checks.
```

tests/testthat/test-tab_afun_cfun.R
Melkiades: I still guess you entered the function from the browser() and it was missing, so probably it was triggering the error across the loop? Or how did you get the error?

anajens: I put browser() inside `cfun_tmp`.  Here's a better reprex:

```
ADSL <- ex_adsl

ccfun <- function(df, labelstr, .N_col, .alt_df) {
    in_rows(
        rcell(nrow(df) * c(1, 1 / .N_col), format = "xx (xx.xx%)"),
        .labels = labelstr
    )
}

lyt <- basic_table(show_colcounts = TRUE) %>%
    split_rows_by("STRATA1", split_fun = keep_split_levels("A")) %>% 
    summarize_row_groups(cfun = ccfun)  %>%
    analyze("BMRKR1")

build_table(lyt, ADSL) # see warning
build_table(lyt, ADSL, alt_counts_df = DM) # no warning

```

I only see the warning in the case of `build_table` without `alt_counts_df` and the cfun includes `.alt_df`.  It's odd because the warning persists once it pops up and then disappears  :
image

R/tt_dotabulation.R
```diff
@@ -1018,10 +1116,11 @@ build_table <- function(lyt, df,
                            topleft)
       if(!is.null(col_counts))
         disp_ccounts(cinfo) <- TRUE
-
+
```
gmbecker, line 1119: whitespace

tests/testthat/test-tab_afun_cfun.R
gmbecker: I think we should mention columns rather than simplifying all the way down to what @anajens mentioned, because it is indicating where the problem occurred. 

Other than that, honestly these checks should be happening at build_table time rather than in the guts of tabulation, I think. For now its probably fine though.

tests/testthat/test-tab_afun_cfun.R
gmbecker: This looks like a paste error where we are constructing the message but one of the elements in inexpectedly of length `> 1`

gmbecker (changes requested):

tests/testthat/test-tab_afun_cfun.R
Melkiades:
```r
> build_table(lyt, ADSL) # see warning
Called from: parent_cfun(df = df, .N_col = .N_col, .something = .something, 
    labelstr = clabelstr, ...)
Browse[1]> .something
Error in (function (df, .N_col, .something, ...)  : 
  argument ".something" is missing, with no default
In addition: Warning messages:
1: In get(object, envir = currentEnv, inherits = TRUE) :
  restarting interrupted promise evaluation
2: restarting interrupted promise evaluation 
Browse[1]> labelstr
[1] "A"
Warning message:
In get(object, envir = currentEnv, inherits = TRUE) :
  restarting interrupted promise evaluation
Browse[1]> Q
```
This happens also on main with .something

R/colby_constructors.R
```diff
@@ -871,11 +883,63 @@ split_rows_by_cutfun <- function(lyt, var,
...
+#'   \item{.ref_full}{data.frame or vector of subset corresponding to the
+#'     `ref_group` column without subsetting defined by row-splitting. Optional
+#'     and only required/meaningful if a `ref_group` column has been defined}
+#'   \item{.in_ref_col}{boolean indicates if calculation is done for cells
+#'     within the reference column}
+#'   \item{.spl_context}{data.frame, each row gives information about a
+#'     previous/'ancestor' split state. See \code{\link{spl_context}}}
+#'   \item{.alt_df}{data.frame, i.e. the `alt_count_df` after 
+#'     row splitting. It can be used with `.all_col_exprs` and `.spl_context`
+#'     information to retrieve current faceting, but for `alt_count_df`.}
```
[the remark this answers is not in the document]
Melkiades: It is producing an error that informs you that it is the split of `alt_counts_df` that failed. I will make it more explicit in docs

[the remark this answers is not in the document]
Melkiades: I thought about this and put a warning but I think it is not necessary. The documentation needed to highlight this nonetheless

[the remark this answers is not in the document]
Melkiades: The warning is for something different (i.e. if the filtering by rows and/or cols gives nothing). After discussing yesterday, I think this is not supposed to be a warning, if you filter out all the variables, it is fine to have empty tables (also because it is always used for calculating counts, for which it makes sense to have none).

R/tt_dotabulation.R
```diff
@@ -2,69 +2,107 @@
 match_extra_args <- function(f,
                             .N_col,
                             .N_total,
+                            .all_col_exprs, 
+                            .all_col_counts,
```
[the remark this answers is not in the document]
Melkiades: It is the nrows of alt_counts_df from all the columns (so you can access them). The impossible split is considered an error and informatively sent back if it happens (there is `.alt_df` and `.alt_df_row`). Otherwise only df counts are used

R/tt_dotabulation.R
```diff
@@ -2,69 +2,107 @@
...
                      takesdf = .takes_df(func),
                      baselinedf,
+                     alt_dfpart,
                      inclNAs,
                      col_parent_inds,
                      spl_context) {
-
-    spl_context$cur_col_subset <- col_parent_inds
-    spl_context$cur_col_n <- vapply(col_parent_inds, sum, 1L)
+    if (NROW(spl_context) > 0) {
```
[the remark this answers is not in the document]
Melkiades: It was breaking some of the assignments afterwards, I do not remember when but there are cases in which some of the relevant spl_contet columns are not assigned. I think this is safer than doing only a couple of values this way

R/tt_dotabulation.R
```diff
@@ -668,26 +722,53 @@ setMethod(".make_split_kids", "Split",
...
+    }
 
 
-    stopifnot(length(newbaselines) == length(dataspl),
-              length(newbaselines) == 0 ||
-                  identical(unique(sapply(newbaselines, length)),
-                            length(col_exprs(cinfo))))
     innerlev <- lvl + (have_controws || is.na(make_lrow) || make_lrow)
-
+
```
Melkiades, line 769:
```suggestion
```

R/tt_dotabulation.R
```diff
@@ -1018,10 +1112,11 @@ build_table <- function(lyt, df,
                            topleft)
       if(!is.null(col_counts))
         disp_ccounts(cinfo) <- TRUE
-
+
```
Melkiades, line 1115:
```suggestion
```

tests/testthat/test-tab_afun_cfun.R
Melkiades: I did not fix the double error print as it seems a problem unrelated to this PR and we can deal with it at a later stage. Changed the ref to dataset name

Melkiades:
@gmbecker updated ;)

Melkiades:
@gmbecker

gmbecker (approved):

Issue #671, opened 2023-07-03
potential issue with whiskers


clarkliming:
bug here

```r
basic_table() %>%
  split_rows_by("Species") %>%
  analyze("Sepal.Length", afun = make_afun(simple_analysis)) %>%
  build_table(iris)

basic_table() %>%
  split_rows_by("Species") %>%
  analyze("Sepal.Length", afun = make_afun(simple_analysis, .labels = list(Mean = "this is {test}"))) %>%
  build_table(iris)

```

this is probably related to reference footnotes, the error message is unclear to the end users. 
```r
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'x' in selecting a method for function 'toString': values must be length 1,
 but FUN(X[[1]]) result is length 0
```

main ref: https://github.com/insightsengineering/chevron/issues/548

gmbecker:
Hi @clarkliming, 

Yes, this is related to referential footnotes. Referential footnotes are encoded in cell text as `{[symbol]}` where `[symbol]` is either the index or a custom symbol.

Pagination happens in formatters now, and formatters only receives the matrix form with already formatted cell values. Pagination involves shuffling referential footnotes around, to determine which page they should be on. Currently, the raw set of referential footnotes is inferred from the matrix form, which involves detecting  `"{*}"` in the cell and label texts. 

For that reason, labels cannot have {} in them.  We can add assertions about this so that the error is more informative.

All that said, I don't understand how this issue would arise from the use of whisker. Even if you're using templating, that template should be populated before the code is run, so that `{test}` in your example should have already been replaced with whatever is going to go there before the `build_table` call is evaluated.

gmbecker:
Your example now gives

```
basic_table() %>%
+   split_rows_by("Species") %>%
+   analyze("Sepal.Length", afun = make_afun(simple_analysis, .labels = list(Mean = "this is {test}"))) %>%
+   build_table(iris)
> 
Error: Error applying analysis function (var - Sepal.Length): Labels cannot contain { or } due to their use for indicating referential footnotes
	occured at (row) path: Species[setosa]
```

as of changes in #673

clarkliming:
the reason is that with whiskers we are introducing some new features in **part** of the table construction. So if our end users is not familiar with this and try to add **whiskers** in other side of the table there can be issues.

gmbecker:
[...]

PR #673, opened 2023-07-03
catch bad labels (with {}) and throw informative error

gmbecker:


gmbecker:
@cicdguy this is failing the roxygen check in the job (though not locally), specifically regarding the `export_as_pdf` function. The issue is that there are no changes to `export_as_pdf` or its roxygen comment in this PR.

Any idea whats happening here (and similarly in other PRs, it seems)?

gmbecker:
> @cicdguy this is failing the roxygen check in the job (though not locally), specifically regarding the `export_as_pdf` function. The issue is that there are no changes to `export_as_pdf` or its roxygen comment in this PR.
> 
> Any idea whats happening here (and similarly in other PRs, it seems)?

So the issue is staged dependencies grabbing updated versions of formatters, and our use of `@inheritParams formatters::export_as_txt` in the documentation to `export_as_pdf`.

Not entirely sure what to do about this, but its an issue the automation team should be aware of, and it'd be great to get some mitigation/flagging if possible.

shajoezhu:
> @cicdguy this is failing the roxygen check in the job (though not locally), specifically regarding the export_as_pdf function. The issue is that there are no changes to export_as_pdf or its roxygen comment in this PR.

Roxygen issue should be fixed now

shajoezhu (approved):
thanks a lot Gabe! the message is more informatvie

shajoezhu:
close #671

PR #676, opened 2023-07-05
670 rename as ard@main

Nolan-Steed:
Let me know if any further changes are needed, thanks!


Melkiades (changes requested):
Hi Nolan! Thanks for this PR. Note that we do not have yet a general style enforced in `rtables` so modifying it partially can be very problematic. Could you revert all the style changes? Also, I suggest you start by renaming files (in rstudio can be done) so that it knows they are actually the same file and does not remove-add everything making it impossible to review

Nolan-Steed:
Made fixes in #677

PR #677, opened 2023-07-06
670 rename as ard 2@main

### LINKED ISSUE #670: rename function "as_ard" to "as_dataframe_with_spec"
hey guys @gmbecker @Nolan-Steed , I would like to suggest renaming this function “as_ard", given ARD is a much larger topic and yet to be defined. I think this is a very great first step, but might be misleading later on. I would suggest to rename this function for now. Thanks!

cc @telepath37 @pawelru @khatril for awareness

Nolan-Steed:
Resolves #670


Melkiades (commented):
LGTM! Thanks Nolan :) 

@gmbecker @shajoezhu

shajoezhu (approved):
lgtm! Thanks @Nolan-Steed @Melkiades

gmbecker (changes requested):
We can change the name if we need but I want to workshop what the new name will be a bit more. 

I don't think flattened dataframe is specific enough, because the path enriched df is also flat.

I'll propose a naming scheme here soon

Nolan-Steed:
@gmbecker sounds good, I will update the names once you come up with them

Nolan-Steed:
Hi @gmbecker if you could send your proposed names within the next day or two so that I can update the code so we could add this in the July release. Thank you for your help!

shajoezhu:
lgtm! Thanks @Nolan-Steed 

@gmbecker , can you please take another look at this, and approve the PR. Thanks

gmbecker (approved):

PR #680, opened 2023-07-11
Integrate support for decimal alignment tests

### LINKED ISSUE #610: Allow decimal alignment coming from formatters
Depends on https://github.com/insightsengineering/formatters/pull/127

Melkiades:
Fixes #610


Melkiades:
@gmbecker Can we merge this?

NEWS.md
```diff
@@ -1,5 +1,6 @@
 ## rtables 0.6.2.9001
  * Cleaned up spelling in documentation ([#685](https://github.com/insightsengineering/rtables/issues/685))
+ * Added decimal alignment test. Custom appearance vignette and alignment checks have been updated with modifications coming from `formatters`.
```
gmbecker, line 3: I don't think we should mention added tests in the NEWS file, please edit this so it just mentions the custom appearance vignette and alignment checks (either as one entry or two) and be a little more detailed about those changes

R/utils.R
```diff
@@ -101,12 +102,11 @@ is_logical_vector_modif <- function(x, min_length = 1) {
...
 #' @export
 rtables_aligns <- function() {
-  c("left", "right", "center")
+  list_valid_aligns()
 }
 
-chk_rtables_align <- function(algn) {
-    if(any(is.na(algn) | !(algn %in% rtables_aligns())))
-        stop("Unsupported text-alignment: ", algn)
+chk_rtables_align <- function(algn) { # xxx duplication with formatters
```
gmbecker, line 108: `check_aligns` is exported from formatters, lets just refactor to use that directly rather than doing this (as your comment correctly points out) silly duplication.

R/utils.R
```diff
@@ -101,12 +102,11 @@ is_logical_vector_modif <- function(x, min_length = 1) {
 #'
 #' @export
 rtables_aligns <- function() {
-  c("left", "right", "center")
+  list_valid_aligns()
```
gmbecker, line 105: deprecate this in favor of `formatters::list_valid_aligns`

vignettes/custom_appearance.Rmd
```diff
@@ -37,7 +37,8 @@ library(dplyr)
 
 It is possible to align the content by assigning `"left"`, `"center"`
 (default), and `"right"` to `.aligns` and `align` arguments in
-`in_rows()` and `rcell()`, respectively.
+`in_rows()` and `rcell()`, respectively. It is also possible to use `decimal`,
```
gmbecker, line 40: We should take a sentence or two to define what decimal alignment means, and also either incorporate it into one of the existing examples in this vignette, or add a new one which showcases it.

R/utils.R
gmbecker: We should find out of this breaks anything downstream, if so, if so it should be deprecated right after the august release, if not we should deprecate before the release.

gmbecker (changes requested):

R/utils.R
Melkiades: I checked. It is never used downstream as far as I could find with staged.deps and manual search

gmbecker (approved):

Issue #651, opened 2023-06-08
[Bug]: split_rows_by when setting both (split_fun = drop_split_levels and page_by = TRUE) for empty table fails


clarkliming:
### What happened?

split_rows_by with split_fun = drop_split_levels and page_by = TRUE for empty table fails with error

Error in TableTree(kids = inner, name = obj_name(spl), labelrow = LabelRow(label = obj_label(spl),  : 
  Got a page title prefix for an Elementary Table

### sessionInfo()

```R
R version 4.2.2 Patched (2022-11-10 r83330)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.5 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/blas/libblas.so.3.9.0
LAPACK: /usr/lib/x86_64-linux-gnu/lapack/liblapack.so.3.9.0

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] scda_0.1.5.9005    tern_0.8.2.9006    rtables_0.6.1.9002 magrittr_2.0.3    
[5] formatters_0.5.0  

loaded via a namespace (and not attached):
 [1] pillar_1.8.1     compiler_4.2.2   forcats_1.0.0    tools_4.2.2     
 [5] digest_0.6.31    jsonlite_1.8.4   lifecycle_1.0.3  tibble_3.1.8    
 [9] gtable_0.3.1     checkmate_2.1.0  lattice_0.20-45  pkgconfig_2.0.3 
[13] rlang_1.0.6      Matrix_1.5-3     cli_3.6.0        fastmap_1.1.0   
[17] withr_2.5.0      dplyr_1.1.0      generics_0.1.3   vctrs_0.5.2     
[21] grid_4.2.2       tidyselect_1.2.0 glue_1.6.2       R6_2.5.1        
[...]

PR #686, opened 2023-07-13
Possible solution with a warning to `page_by` nothing

Melkiades:
Disclaimer: I largely agree with Gabe on this and it is probably better to check this information beforehand, but an informative error/warning is needed imo and it cannot be thrown before the split sees the data (where I put the check). I was already working on it so I am posting it anyway. I am open to any consideration :)

#651


R/tt_dotabulation.R
```diff
@@ -720,6 +719,18 @@ setMethod(".make_split_kids", "Split",
     SIMPLIFY = FALSE))
 
     inner <- .set_kids_sect_sep(inner, spl)
+    
+    # Trying to split nothing with page_by
+    if (is.null(inner) && !is.na(ptitle_prefix(spl))) {
```
gmbecker, line 724: The issue here is that there are going to be expectations about tables that come out of applying a layout that has page-by splitting. Most importantly, that there will be page titles for any table that is returned an eventually exported. QC is likely to (correctly) look for them, etc. The tables are "wrong" from the user perspective without them. 

This returns a table that isn't going to be the correct shape, and I am still very much not convinced that is better than returning no table at all and throwing an error. In fact I'm pretty strongly convinced of the opposite.

Melkiades: I agree with having the error. Also, later the pagination error for it is very mysterious. That would be solved with sanitization though

shajoezhu:
closing this PR, as there is a new implementation in https://github.com/insightsengineering/rtables/pull/712

PR #700, opened 2023-07-27
Fixes to various row_labels situations with tests.

### LINKED ISSUE #698: [Bug]: row_labels don't match docs in qtable()
### What happened?

According to the docs, `row_labels` from `qtable()` should be used to modify the analysis rows. In practice it seems to only change the top left header label.

### Relevant log output
```
fivenum3 <- function(x) {
    as.list(fivenum(x))
}
# annotate top left
qtable(ex_adsl, row_vars = "STRATA2", col_vars = "ARM", avar = "AGE", afun = fivenum3,
       row_labels = "ABC")

# error
qtable(ex_adsl, row_vars = "STRATA2", col_vars = "ARM", avar = "AGE", afun = fivenum3,
       row_labels = letters[1:5])
```

```
ABC   A: Drug X   B: Placebo   C: Combination
       (N=134)     (N=134)        (N=132)    
—————————————————————————————————————————————
S1                                           
        23.00       24.00          20.00     
        28.00       30.00          30.50     
        34.00       36.00          35.00     
        39.00       40.50          40.00     
        48.00       62.00          69.00     
S2                                           
        21.00       21.00          23.00     
        29.00       29.50          30.00     
        32.00       32.00          34.50     
        38.00       39.50          38.00     
        50.00       58.00          64.00     
> 
> # error
> qtable(ex_adsl, row_vars = "STRATA2", col_vars = "ARM", avar = "AGE", afun = fivenum3,
+        row_labels = letters[1:5])
Error: Error applying analysis function (var - AGE): 'length = 5' in coercion to 'logical(1)'
occured at (row) path: STRATA2[S1]
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

gmbecker:
Close #698.


anajens (changes requested):
Thanks @gmbecker. Here is my feedback on some minor things:  

1. I think it would be better if the analysis variable and analysis function are always annotated regardless of the type of table variants {1-4} as described in #698. 
If users think the default label is too long it can easily be changed with `row_labels`.

For example:
```
qtable(ex_adsl , col_vars = "ARM")
```

Should return
```
                   A: Drug X   B: Placebo   C: Combination
                    (N=134)     (N=134)        (N=132)    
——————————————————————————————————————————————————————————
STUDYID - count      134         134            132 
```


3. What do you think about changing the default formatting of `avar - afun` label to `afun(avar)`. For example, instead of `AGE - max` to `max(AGE)` ? Given that the purpose is exploratory, this would be a nice reminder to the person doing the analysis how the table calculations are done.

4. This should throw an error as the `row_labels` length shouldn't exceed 1 for table variant {2}:
```
qtable(ex_adsl , row_vars = "STRATA2", col_vars = "ARM", avar = "AGE", afun = mean, 
       row_labels = c("ABC", "EFG", "HIJ"))
ABC                                          
EFG   A: Drug X   B: Placebo   C: Combination
HIJ    (N=134)     (N=134)        (N=132)    
—————————————————————————————————————————————
S1      34.10       36.46          35.70     
S2      33.38       34.40          35.24     
```

5. For multi-row returning afun the length of `row_labels` should match the number of cells created by `afun`. Right now nothing happens but I think this should also throw an error. Same thing happens in the case of this table with the additional `row_vars` split:
```
qtable(ex_adsl, col_vars = "ARM", avar = "AGE", afun = fivenum3,
       row_labels = "ABC") 
AGE - fivenum3   A: Drug X   B: Placebo   C: Combination
                  (N=134)     (N=134)        (N=132)    
————————————————————————————————————————————————————————
                   21.00       21.00          20.00     
                   28.00       30.00          30.00     
                   33.00       35.00          35.00     
                   39.00       40.00          40.00     
                   50.00       62.00          69.00   
```

gmbecker:
@anajens I didn't change the label display but both directions of row_labels - number of afun cells mismatch are now errors.

anajens:
> Also there are now separate `qtable_layout` and `qtable` functions, as we discussed.

Thanks! I think this is a good start but not exactly what I had in mind. I was thinking about being able to return the expression of the layout so users can see directly the connection from qtable to the layout framework. But that's probably a much bigger task for another day ( if users request this).

gmbecker:
> > Also there are now separate `qtable_layout` and `qtable` functions, as we discussed.
> 
> Thanks! I think this is a good start but not exactly what I had in mind. I was thinking about being able to return the expression of the layout so users can see directly the connection from qtable to the layout framework. But that's probably a much bigger task for another day ( if users request this).

That is possible to do but out of scope for this issue, as it would require non-trivial new development. It is a good idea though

anajens:
New issue added: https://github.com/insightsengineering/rtables/issues/711

PR #721, opened 2023-08-23
Catching errors for mismatches in alt_counts_df splits

### LINKED ISSUE #720: Possible bug for `alt_counts_df` split by `NA_character_` or different number of levels
We need to test these cases and catch bugs

```r
    afun_tmp <- function(x, .alt_df_row, ...) mean(x)
    lyt_row <- basic_table() %>% split_rows_by("ARMCD") %>% analyze("BMRKR1", afun = afun_tmp)
    DM_tmp <- DM %>% mutate("ARMCD" = NA_character_)
    lyt_row %>% build_table(ex_adsl, alt_counts_df = DM_tmp)
```

Melkiades:
Fixes #720


R/tt_dotabulation.R
```diff
@@ -759,6 +759,18 @@ setMethod(".make_split_kids", "Split",
                  alt_dfpart$message,
                  call. = FALSE)
         }
+        
+        # Error if split does not have the same values in the alt_df (and order)
+        alt_df_spl_vals <- levels(as.factor(alt_df[[spl_payload(spl)]]))
+        # The following breaks if there are different levels (do_split returns empty list)
+        # or if there are different number of the same levels
+        if (length(alt_dfpart) != length(dataspl)) {
```
gmbecker, line 767: I think we can (and should) make the requirement stronger. We don't just want the same number of facets, we want _**the same facets**_. Check if the sets of values are actually the same, rather than just having the same count.

gmbecker (changes requested):

R/tt_dotabulation.R
Melkiades: If the data does not contain the split values contained in the split (`spl`) which are defined in `df`, then the `do_split` on `alt_df` produces an empty list, giving a different length of obj. I can try to see what happens if there is a mix...

Melkiades: covered!

Melkiades:
@gmbecker @shajoezhu, I think this is important to get in before CRAN release

shajoezhu (commented):
hi @Melkiades thanks for the change, it looks much better

i am getting more informative errors now

```
Error in .local(spl, have_controws, make_lrow, ...) :
alt_counts_df split variable(s) [ARMCD] (in split VarLevelSplit) has not the same factor levels of df.
df has c(ARM A, ARM B, ARM C) levels while alt_counts_df has c()
```

can you add quotes to the arms, so it appears like `c("ARM A", "ARM B", "ARM C") `

Thanks a lot

Melkiades:
> hi @Melkiades , thanks for the change
> 
> looks better, but still need change. I am getting
> 
> ```
> Error in .local(spl, have_controws, make_lrow, ...) :
> alt_counts_df split variable(s) [ARMCD] (in split VarLevelSplit) has not the same factor levels of df.
> df has c("ARM A", "ARM B", "ARM C") levels while alt_counts_df has c("NA")
> > levels(DM_tmp$ARMCD)
> NULL
> ```
> 
> `c("NA")` is not right in this case

it is not a factor, so it has no levels

PR #739, opened 2023-09-26
Fix bug causing all-`NA` rows to be included in all `.df_row` splits

### LINKED ISSUE #687: [Bug]: `.df_row` includes rows with row split var `NA` values in all split levels
When splitting on a variable which has `NA` values, the rows with these `NA` values are included in `.df_row` for all split levels.

For example,

```r
library(rtables)

a = data.frame(
  a = factor(c('a', NA,  'b', 'b', 'a', 'a', 'b', 'c', 'a', NA)),
  b = factor(c(NA,  NA,  'x', 'x', 'y', 'x', 'x', 'y', 'x', NA))
)

basic_table() %>%
  add_overall_col("all pts") %>%
  split_rows_by("a") %>%
  analyze(
    "b",
    afun = function(x, .df_row, ...) in_rows(.list = list("nrow(.df_row)" = nrow(.df_row)))  
  ) %>%
  build_table(a)
```
produces
```
                  all pts
—————————————————————————
a                        
  nrow(.df_row)      6   
b                        
  nrow(.df_row)      5   
c                        
  nrow(.df_row)      3   
```

edelarua:
Closes #687 

No downstream changes.


Melkiades (approved):
Amazing catch Emily!!! This is very good, I tried with our two examples and also modified them to try to have the split with NA only on the split var but of course, it has been trimmed before. I think as a general rule for the future, as I told a couple of times, I would always keep NA as unique missing indicator and if someone wants to pivot over it, they need to make it an explicit factor on their own account. Anyway `inclNA` is not something we should support too much as it has a lot of edge cases that make little sense and have basically no uses. I wounder if this will affect counts for alt_counts_df downstream.... mmmh probably not because already the splits already take out NA values from the count. What do you think @edelarua @ayogasekaram?

PR #748, opened 2023-10-16
update ref_group column ordering

### LINKED ISSUE #747: [Question]: How to make a custom split function to place reference group last
### What is your question?

Followup to https://github.com/insightsengineering/rtables/issues/707

I'm trying to write a custom split function to place the reference group last but I'm stuck. 

```
ref_last <- function(splret, spl, fulldf, ...) {

    cat("Initial: ", names(splret$values), "\n")

    init_ord <- 1:length(splret$values)

    in_ref_col <- vapply(splret$values, function(x){x@extra$.in_ref_col}, logical(1))
    ref_group_pos <- init_ord[in_ref_col]

    # Put the reference group at the end
    ord <- c(init_ord[-ref_group_pos], init_ord[ref_group_pos])

    cat("Reorder:", names(splret$values)[ord], "\n")

    make_split_result(splret$values[ord],
                    splret$datasplit[ord],
                    splret$labels[ord])
}
```

Here is the example:
```
# levels of x are not alphabetical in order
dat2 <- data.frame(
  x = factor(letters[1:5], levels = letters[5:1]),
  y = 1:5
)

> # Expect to mainain order as "e" "d" "c" "b" "a" since ref group is already last
> basic_table() %>%
+   split_cols_by("x", ref_group = "a", split_fun = make_split_fun(post = list(ref_last))) %>%
+   analyze("y") %>%
+   build_table(dat2)
Initial:  e d c b a 
Reorder: e d c b a 
Initial:  a b c d e 
Reorder: b c d e a 
        b      c      d      e      a  
———————————————————————————————————————
Mean   2.00   3.00   4.00   5.00   1.00
```

Why is the split_fun called twice? The second time reordering seems to return the incorrect result.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct.

### Contribution Guidelines

- [X] I agree to follow this project's Contribution Guidelines.

### Security Policy

- [X] I agree to follow this project's Security Policy.

clarkliming:
close #747


shajoezhu:
test snap changing at https://github.com/insightsengineering/scda.test/pull/67

anajens (commented):
I think this is too big of a breaking change right before the release. All of tern and TLG-C examples using `ref_group` will be affected.

Ideally the `ref_group` column should continue to be placed first by default. We just need to offer support to users who do not wish to use the default.

clarkliming:
after simple evaluation, the `ref_group` is not affecting many templates. 12 templates in tlg-catalogue contains the `ref_group` argument, and only one snapshot is affected (ratet01, using ARM B as the reference group). Others all using the first level of arm as reference (so not changing snapshots).

Honestly I think this change should happen much earlier. With this update more flexibility is introduced and we finally have a consistent way of working with columns. Now factor levels is the only thing that changes the column order.

Melkiades:
> after simple evaluation, the `ref_group` is not affecting many templates. 12 templates in tlg-catalogue contains the `ref_group` argument, and only one snapshot is affected (ratet01, using ARM B as the reference group). Others all using the first level of arm as reference (so not changing snapshots).
> 
> Honestly I think this change should happen much earlier. With this update more flexibility is introduced and we finally have a consistent way of working with columns. Now factor levels is the only thing that changes the column order.

I understand your point of view, but I think it is better to keep the standard like it is and then have a split_fun in tern that takes care of any possible change. Here, we added 3 ways to do this https://github.com/insightsengineering/tern/pull/1087/files. @anajens mentioned the possibility of having a lot of columns, in that case having the ref group as standard would be very useful, but I do not know if this is in line with the standards (also last is the same practically). Maybe @barnett11 can help us in this issue!

barnett11:
Thanks @Melkiades, I would expect Control group to be either first or last, but fixing it one place could be difficult. The standards do not mandate that control arm must come first, so i think we should allow flexibility for this change in order as we do for all of our templates. It is common, and STREAM has allowed this flexibility to users for some time

Melkiades:
Just added blocked because I want to add the split functions at the same time so not to have changes in the TLG-c

Melkiades:
btw changing ref_group positioning is affecting all of these, it was not apparent because the ref_group is by chance the first group: Screenshot 2023-10-27 173235

Melkiades (approved):

PR #771, opened 2023-11-01
Render the dev-guide as part of pkgdown

cicdguy:
Instead of storing pre-rendered HTML and static assets in the repo.
Also removed a bunch of trailing spaces.


edelarua (approved):
Lgtm - thanks Dinakar!

PR #772, opened 2023-11-01
Fix bug when footnote path contains named elements

edelarua:
Part of https://github.com/insightsengineering/tern/issues/1083

Required for https://github.com/insightsengineering/tern/pull/1116


Melkiades (approved):
I think this makes sense. Can you point me out in {tern} exactly where this was coming up?

edelarua:
If you run the refactored `surv_time` function on [`1083_surv_time_range@main`](https://github.com/insightsengineering/tern/tree/1083_surv_time_range%40main) you will get an error when generating cell footnotes which traces back to this line. I believe this is the first `tern` function where we've tried to add in cell footnotes which is why we haven't encountered this before.

Melkiades:
Thanks Emily!! Amazing catch here :)

PR #777, opened 2023-11-07
Keep whitespace in html tables

### LINKED ISSUE #775: white spaces are trimmed by html
while we keep the white spaces in text, in the html output they are discarded.

Suggest to use "&nbsp" to replace the leading whitespaces so that the white spaces are kept

Or use some CSS style like `white-space: pre;` to prevent it

example css
```{CSS}
<style>
    td {white-space: pre;}
</style>
```

edelarua:
Closes #775


Melkiades (approved):
Lgtm! Can you add a small test? Thanks :)

PR #781, opened 2023-11-09
Combine duplicate ref footnotes

### LINKED ISSUE #779: Combine duplicate referential footnotes
In many cases, the same referential footnote is used in multiple places. It would be a lot neater and easier to read if each unique referential footnote was only listed once and ref index numbers were instead repeated within tables.

Example (taken from `test-header-footer.R`):

```
Title says Whaaaat
Oh, ok.

——————————————————————————————————————————————————
           A: Drug X   B: Placebo   C: Combination
——————————————————————————————————————————————————
row1 {1}       5           5              5       
row2       1, 2 {2}     1, 2 {3}       1, 2 {4}   
——————————————————————————————————————————————————

{1} - row 1 rfn
{2} - row 2 cfn
{3} - row 2 cfn
{4} - row 2 cfn
——————————————————————————————————————————————————

ha HA! Footer!
```

changes to:

```
Title says Whaaaat
Oh, ok.

——————————————————————————————————————————————————
           A: Drug X   B: Placebo   C: Combination
——————————————————————————————————————————————————
row1 {1}       5           5              5       
row2       1, 2 {2}     1, 2 {2}       1, 2 {2}   
——————————————————————————————————————————————————

{1} - row 1 rfn
{2} - row 2 cfn
——————————————————————————————————————————————————

ha HA! Footer!
```

edelarua:
Closes #779 

Merge after https://github.com/insightsengineering/formatters/pull/225


R/tt_toString.R
```diff
@@ -435,7 +435,6 @@ get_formatted_fnotes <- function(tt) {
   )
 
   inds <- vapply(lst, ref_index, 1L)
-  stopifnot(all(is.na(inds)) || !is.unsorted(inds))
```
Melkiades, line 438: why is this no more valid?

R/tree_accessors.R
```diff
@@ -2904,6 +2904,23 @@ setMethod("col_fnotes_here<-", "LayoutColLeaf", function(obj, value) {
   obj
 })
 
+#' @export
+#' @rdname int_methods
+setMethod(
+  "col_fnotes_here", "VTableTree",
```
Melkiades, line 2910: This extracts all fnotes right?

edelarua: This is for column referential footnotes - we also have `row_footnotes()` for row ref footnotes and `cell_footnotes()` for cell ref footnotes.

R/tt_toString.R
edelarua: It's not required that indices are sorted after this change - for example, if you have footnotes 1, 2, and 3, and footnote 1 is repeated in a later cell, you might have indices ordered [1, 2, 3, 1], which fails this check.

edelarua: To clarify, I left all footnotes as separate entries in `ref_fnote_df` from `matrix_form` and just changed the symbol/ref_index columns. The symbol info seems to be arbitrary as I can't find anywhere that it's used. Then in formatters, I removed the auto-reindexing (`fix_fnote_df` which overrides indices to 1, 2, 3, ...), and set the `toString` method to only print each unique ref footnote once.

R/tree_accessors.R
edelarua: Yes they are exported. We should be able to rename/deprecate - I highly doubt anyone has used these functions outside of the package

R/tree_accessors.R
```diff
@@ -2877,33 +2877,70 @@ setMethod("cell_footnotes<-", "ContentRow",
...
+  col_footnotes(obj)
+})
+#' @export
+#' @rdname ref_fnotes
+setGeneric("col_fnotes_here<-", function(obj, value) standardGeneric("col_fnotes_here<-"))
+#' @export
+#' @rdname int_methods
+setMethod("col_fnotes_here<-", "ANY", function(obj, value) {
+  warning("col_fnotes_here<- is deprecated since {rtables} version 0.6.5.9011. Please use col_footnotes<- instead.")
+  col_footnotes(obj) <- value
```
Melkiades, line 2897: @edelarua is this a valid deprecation cycle, what do you think?

edelarua: Looks good to me!

R/tree_accessors.R
```diff
@@ -2877,33 +2877,70 @@ setMethod("cell_footnotes<-", "ContentRow",
...
+  warning("col_fnotes_here is deprecated since {rtables} version 0.6.5.9011. Please use col_footnotes instead.")
+  col_footnotes(obj)
+})
+#' @export
+#' @rdname ref_fnotes
+setGeneric("col_fnotes_here<-", function(obj, value) standardGeneric("col_fnotes_here<-"))
+#' @export
+#' @rdname int_methods
+setMethod("col_fnotes_here<-", "ANY", function(obj, value) {
+  warning("col_fnotes_here<- is deprecated since {rtables} version 0.6.5.9011. Please use col_footnotes<- instead.")
```
Melkiades, line 2896: I think it is better to use .Deprecated

Melkiades (approved):
Lgtm Emily!! Great work as always :)

R/tree_accessors.R
edelarua: I'll switch it to that! I don't think it makes too much of a difference either way

R/tree_accessors.R
```diff
@@ -2877,33 +2877,83 @@ setMethod("cell_footnotes<-", "ContentRow",
...
+#' @export
 #' @rdname int_methods
-setMethod("col_fnotes_here", c("LayoutColTree"), function(obj) obj@col_footnotes)
+setMethod("col_fnotes_here<-", "ANY", function(obj, value) {
+  .Deprecated(
+    new = "col_footnotes<-",
+    old = "col_fnotes_here<-",
+    msg = "col_fnotes_here<- was deprecated in {rtables} version 0.6.5.9011. Please use col_footnotes<- instead."
+  )
+  col_footnotes(obj) <- value
```
Melkiades, line 2905: only a silly question: why dont we need here to have obj at the end while in other places we do?

edelarua: I'm not sure, possibly accidental? I don't think these functions were ever fully fleshed out

PR #794, opened 2023-11-29
Fix potential tt at path issue

clarkliming:
there is a fix for path names
https://github.com/insightsengineering/rtables/blob/478e637a7ca2912c57c3b8f2532375c906a90331/R/index_footnotes.R#L127

however tt_at_path is still not updated. It is not robust enough.

```
library(tern)
library(dplyr)
adtte_f <- tern_ex_adtte %>%
  filter(PARAMCD == "OS") %>%
  mutate(
    AVAL = day2month(AVAL),
    is_event = CNSR == 0
  )

adtte_f$is_event <- FALSE

tbl <- basic_table() %>%
  split_cols_by(var = "ARMCD", ref_group = "ARM A") %>%
  add_colcounts() %>%
  surv_time(
    vars = "AVAL",
    var_labels = "Survival Time (Months)",
    is_event = "is_event",
    control = control_surv_time()
  ) %>%
  build_table(df = adtte_f)

tt_at_path(tbl, row_paths(tbl)[[5]])
```

it will give you an error that the path is not valid.

Let's fix this issue from the root. Let's not use "identical" for comparison (to avoid potential issue of character names), or let's just use unname() for all occurences of row_paths, tt_at_path, etc. ideally the row_paths should not have any name.


Melkiades (approved):
Thanks Liming for finding this bug! I could not reproduce it in base {rtables} but I corrected the more general `row_paths()` that is based on `make_rows_df()` so now it will not produce named paths. I checked with `surv_time` and `cell_values` which was also breaking and the latter fix works. I think it is fine not to have `identical` as the label is not really a needed check. I added a test but it is not generating named paths. @edelarua do you know how it is possible to correct the test to have it doing what `surv_time` does?

Issue #785, opened 2023-11-15
Support core split overriding split functions in column space


gmbecker:
This will involve refactoring how how, and more importantly where, column subset expressions are generated. 

@Melkiades @edelarua I'm not able to assign labels currently but please create "gabe" and "sme_review" labels (or similar) and apply them here.  Also apply whatever I called the "collaborator request" type label.

I know how to do this and will implement it.

shajoezhu:
Hi @gmbecker , can you please help me to better understand the rationale for this request? Thanks

gmbecker:
@shajoezhu Certain complex tables apparently require columns along the lines of "patients enrolled and then so and so happened later", which is a subset of "patients enrolled". And these subsets can be overlapping, or interacting in other ways.

As I understand it, its something along the lines of:

```

         All Patients              Pats With A      Pats w/ A and B       Pats w/ B and not A 
----------------------------------------------------------------------------------------------
n            75                        50                 30                      15
mean                                            ...
```

etc etc. 

We have something that fully supports this in row space, but it currently isn't supported in column space because of differences in how the tabulation machinery handles row and column subsetting (the former is done by subsetting the actual dataframe, the latter is carried around as unevaluated expressions).

I have something working locally in my fork that is passing all the direct rtables tests including a new one for this functionality.

Melkiades:
But if you do `split_cols_by` and then add a split function that does this?

gmbecker:
@Melkiades it depends on the split function. If you use the provided combo functionality (e.g., add_combo_facet), it will work, but general custom splitting functions, specifically ones that override core splitting, are disallowed in column space. 

You can see that from the current tests:

https://github.com/insightsengineering/rtables/blob/4a445e72e303a403b2ac7ab5585ce14f211b2f7a/tests/testthat/test-split_funs.R#L353-L358

That is what I have fixed locally by changing how column subsetting expressions are handled. I'll create a PR, probably today, based on my fork.

gmbecker:
Sorry for the delay, this got buried in other things and then lost over the thanksgiving break in america. Anyway, PR is up now

gmbecker:
This was closed by #875

PR #795, opened 2023-11-29
refactor colsubset exprs tracking so it an be overridden in split funs

gmbecker:
This closes insightsengineering/rtables#785, allowing custom split functions, specifically functions which override the `core` function, to specify subsetting expressions for the facets they generate, which is necessary for the custom split to function properly in column space. 

This is backwards compatible other than the fact that attempting to use a custom splitting function which overrides core splitting behavior in column space was previously an error, and any TableTree objects that have been serialized to rds/rda files would need to be refreshed as one of the underlying classes has changed. Tests have been added/updated for the new behavior

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rtables
 Title: Reporting Tables
-Version: 0.6.6.9004
-Date: 2024-01-17
+Version: 0.6.6.9005
+Date: 2024-01-19
 Authors@R: c(
     person("Gabriel", "Becker", , "gabembecker@gmail.com", role = "aut",
            comment = "Original creator of the package"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,4 +1,4 @@
-## rtables 0.6.6.9004
+## rtables 0.6.6.9005
 ### New Features
  * Added `top_level_section_div` for `basic_table` to set section dividers for top level rows.
 


shajoezhu:
Hi @gmbecker , thanks a lot for the changes. We will need to create PRs and trigger CICD pipelines to investigate if any downstream break changes. Will get back to this as I know more.

R/make_split_fun.R
```diff
@@ -94,27 +96,35 @@ validate_split_result <- function(pinfo, component = NULL) {
...
+#' represent the data associated with facets, while actual data
+#' subsets are used during row splitting.
 #'
 #' @examples
 #' splres <- make_split_result(
 #'   values = c("hi", "lo"),
 #'   datasplit = list(hi = mtcars, lo = mtcars[1:10, ]),
-#'   labels = c("more data", "less data")
+#'   labels = c("more data", "less data"),
+#'   subset_exprs = list(expression(TRUE), expression(seq_along(wt) <= 10))
```
Melkiades, line 111: so this would do a subset only in the column space, while it is ignored in the row space, right?

R/make_split_fun.R
```diff
@@ -124,9 +134,9 @@ make_split_result <- function(values, datasplit, labels, extras = NULL) {
 #' @rdname make_split_result
 #' @param splres list. A list representing the result of splitting.
 #' @export
-add_to_split_result <- function(splres, values, datasplit, labels, extras = NULL) {
+add_to_split_result <- function(splres, values, datasplit, labels, extras = NULL, sub_expr = NULL) {
   validate_split_result(splres)
-  newstuff <- make_split_result(values, datasplit, labels, extras)
+  newstuff <- make_split_result(values, datasplit, labels, extras, subset_exprs = list(sub_expr))
```
Melkiades, line 139: why `sub_expr` is not defaulting to the `splres` values if sub_expr is NULL?

R/make_split_fun.R
```diff
@@ -176,17 +187,17 @@ add_to_split_result <- function(splres, values, datasplit, labels, extras = NULL
...
 #' splitting procedure,  and are only  necessary in rare  cases. These
 #' must  accept  `spl`, `df`,  `vals`,  `labels`,  and can  optionally
-#' accept `.spl_context`. They must return a named list with elements,
-#' all of  the same  length, as follows:  - `datasplit`  (containing a
-#' list of data.frames), -  `values` containing values associated with
-#' the   facets,   which   must   be   `character`   or   `SplitValue`
-#' objects. These  values will  appear in the  paths of  the resulting
-#' table.  - `labels` containing  the character labels associated with
-#' `values`
+#' accept `.spl_context`. They should return an split result object
```
Melkiades, line 193:
```suggestion
#' accept `.spl_context`. They should return a split result object
```

R/make_split_fun.R
```diff
@@ -296,14 +307,14 @@ make_split_fun <- function(pre = list(), core_split = NULL, post = list()) {
 
     if (is.null(core_split)) {
       ret <- do_base_split(spl = spl, df = df, vals = vals, labels = labels)
-    } else if (!in_col_split(.spl_context)) {
+    } else if (TRUE||!in_col_split(.spl_context)) {
```
Melkiades, line 310:
```suggestion
    } else if (TRUE || !in_col_split(.spl_context)) {
```
If there is no stop after the else isnt it better to have this as final else? I think it could be better to add a check for sub_expr instead when `in_col_split(.spl_context)` is true, right?

R/make_split_fun.R
```diff
@@ -356,15 +367,43 @@ make_split_fun <- function(pre = list(), core_split = NULL, post = list()) {
...
+                                    sub_expr = subexpr)
+        add_to_split_result(ret,
+           values = list(val), labels = label,
+           datasplit = datpart
+        )
   }
 }
 
+
+.or_combine_exprs <- function(ex1, ex2) {
```
Melkiades, line 390: I remember us talking about making expression handling more general and contained in a single system of functions. Do you think it is worth it?

Melkiades: I mean adding this to make_subset_expr (or is this too specific?)

R/00tabletrees.R
```diff
@@ -1009,8 +1017,9 @@ make_child_pos <- function(parpos,
     svals = c(pos_splvals(parpos), nsplitval),
     svlabels = c(pos_splval_labels(parpos), newlab),
     sub = .combine_subset_exprs(
-      pos_subset(parpos),
-      make_subset_expr(newspl, nsplitval)
+        pos_subset(parpos),
+        ## this will grab the value's custom subset expression if present
+        make_subset_expr(newspl, nsplitval)
```
Melkiades, line 1022: this is the core effect of altering column col splits

tests/testthat/test_utils.R
```diff
@@ -1,6 +1,7 @@
 context("Checking utility functions")
 
 test_that("func_takes works with different inputs", {
+    func_takes <- rtables:::func_takes
```
Melkiades, line 4: This is not needed as tests access to all the namespace (with hidden functions)
```suggestion
```

Melkiades (commented):
I think this is very good! Thanks Gabe, I will unlock the draft to see if tests sail smoothly ;)

R/make_split_fun.R
gmbecker: Correct. Faceting in column space is tracked via expressions, while faceting in row space is immediately materialized during tabulation. This is largely for a combination of performance and legacy reasons.

R/make_split_fun.R
gmbecker: So there are two cases that can happen here:

1. values are strings (this is more common as this is being called in custom code), in which case value_expr(values[.]) will always be NULL, so that wouldn't change anything. 
2. values are already SplitValue objects, in which case they (should) already have their expressions set (if they are not the default, which will still work in most cases), so they don't need to be changed.

Currently, as written, `subset_expr` ends up being ignored when values are already SplitValue objects.

This is defensible behavior, I think (as the construction time of the SplitValue is where the custom expression "should" be set), but probably not optimal. I'll change things slightly to make it easier to override subset expressions if (for some reason) you're reusing the same value object but want different subsetting behavior (which would be weird but I can think of a a corner case you might want to). As written originally you'd have been expected to reset the expressions on the value objects before passing to make_split_result

R/make_split_fun.R
gmbecker: It should actually just be an else, I'll fix this in a new commit

PR #801, opened 2023-12-06
Fix `as_result_df` bug for small tables

### LINKED ISSUE #800: Fix `as_result_df` bug for small tables
Currently, tables with less than 3 rows return an error when `as_result_df` is applied. Add a default path length of 3 for these tables to fix.

edelarua:
Closes #800


ayogasekaram (approved):
looks good! Thanks @edelarua 
Can we update the news?

PR #811, opened 2024-01-03
`value_formats` addition to vignette and docs

### LINKED ISSUE #598: `value_formats` is stand alone and needs more testing and a note in the vignette
First: should this function be in `formatters`? I was looking for it because I remembered it but it needs to be easier to retrieve than ctrl+f on both reference web pages. The best is to connect it to the rest of the package via links (at least in the section it is in reference). I will open an issue now about that. A vignette description may also put it onto the "user-map".

Just as a reference on how I checked this information regarding `value_formats`:

It is stand-alone:
Screenshot 2023-03-27 184827

It has few and not specific tests:
Screenshot 2023-03-27 184852

It has no vignette mention:
Screenshot 2023-03-27 184921

Melkiades:
Fixes #598


edelarua (approved):
Lgtm! Thanks @Melkiades :)

I just made a couple improvements to the wording of the vignette but otherwise all looks great!

Issue #816, opened 2024-01-22
inconsistent handling of 'root' as first path element and extremely unhelpful error messages in sort_at_path


gmbecker:
tt_at_path (and thus everything it is based on) silently removes 'root' when present as the first element of a path _whether or not that is the actual name of the root TableTree_.

Currently, `sort_at_path` only removes a first path element which matches the name of `tt` (the object having part of it's structure sorted).

This is inconsistent behavior and should be fixed.

Furthermore, this is _**heavily**_ compounded by the fact that there is no useful error signaling from `sort_at_path` when the specified path doesn't exist; it complains about "argument 1 not being a vector", which is actively misleading since the first argument to sort_at_path cannot be a vector, it has to be a `TableTree`

```r
 lyt <- basic_table() %>%
    split_rows_by("SEX", split_fun = keep_split_levels("F")) %>%
    split_rows_by("STRATA1") %>%
    summarize_row_groups() %>%
    analyze("AGE", mean, format = "xx.x")

tbl <- build_table(lyt, ex_adsl)
tbl
```

gives 

```
> tbl
            all obs  
—————————————————————
F                    
  A        63 (15.8%)
    mean      32.4   
  B        73 (18.2%)
    mean      34.8   
  C        86 (21.5%)
    mean      34.4   
```

Then, we have a path that works in `tt_at_path` but not in `sort_at_path`:

```
> tt_at_path(tbl, c("root", "SEX", "F", "STRATA1"))
          all obs  
———————————————————
A        63 (15.8%)
  mean      32.4   
B        73 (18.2%)
  mean      34.8   
C        86 (21.5%)
  mean      34.4   

> sort_at_path(tbl, c("root", "SEX", "F", "STRATA1"), scorefun = cont_n_allcols)
Error in order(scores, na.last = (na.pos != "first"), decreasing = decreasing) : 
  argument 1 is not a vector
```

The associated PR addresses both of these issues so that we get:

```
> sort_at_path(tbl, c("root", "SEX", "F", "STRATA1"), scorefun = cont_n_allcols)
            all obs  
—————————————————————
F                    
  C        86 (21.5%)
    mean      34.4   
  B        73 (18.2%)
    mean      34.8   
  A        63 (15.8%)
    mean      32.4   


> sort_at_path(tbl, c("root", "WRONG", "F", "STRATA1"), cont_n_allcols)
[...]

PR #817, opened 2024-01-23
fix sort behavior for root, add useful error, docs and tests

gmbecker:
Addressses all parts of #816

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -62,7 +62,6 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
 Collate:
     '00tabletrees.R'
     'Viewer.R'
@@ -95,3 +94,4 @@ Collate:
     'tt_from_df.R'
     'validate_table_struct.R'
     'zzz_constants.R'
+RoxygenNote: 7.2.0
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -94,4 +94,4 @@ Collate:
     'tt_from_df.R'
     'validate_table_struct.R'
     'zzz_constants.R'
-RoxygenNote: 7.2.0
+RoxygenNote: 7.3.1
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,7 +1,7 @@
 Package: rtables
 Title: Reporting Tables
-Version: 0.6.6.9005
-Date: 2024-01-19
+Version: 0.6.6.9007
+Date: 2024-02-09
 Authors@R: c(
     person("Gabriel", "Becker", , "gabembecker@gmail.com", role = "aut",
            comment = "Original creator of the package"),
@@ -62,6 +62,7 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
+RoxygenNote: 7.3.1
 Collate:
     '00tabletrees.R'
     'Viewer.R'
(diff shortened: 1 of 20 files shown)


DESCRIPTION
```diff
@@ -62,7 +62,6 @@ Encoding: UTF-8
 Language: en-US
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
```
Melkiades, line 65: I think at the moment, we use 7.3.0

R/tt_sort.R
```diff
@@ -156,7 +180,16 @@ sort_at_path <- function(tt,
   }
 
   ## XXX hacky fix this!!!
+  ## tt_at_path removes root even if actual root table isn't named root, we need to match that behavior
+  if(path[1] == "root") {
```
Melkiades, line 184: I just fixed a problem with `as_result_df` for cases with `rbind`ed tables, and seems that there is a `"rbind_root"` at the beginning of the path. I removed all initial "root" and "rbind_root" from there when doing as_result_df (also because it is not really informative), but I was wondering if we need to check for this also here or if we should fix it in the `rbind` function

gmbecker: I'm not so sure. As I said, I actually think this behavior is outdated and should be deprecated (when the first element, whether root or rbind_root) doesn't actually match the path (ie there is no "root"/"rbind_root" subtable at the root level, its named something else). 

For when it DOES match, it already doesn't matter whether its there or not; you don't have to name the root element when specifying paths (I'm not sure if this is true everywhere... someone should check!, but its true of sort_at_path and tt_at_path, and I think the latter is used by a lot of the user-facing pathing support).

Put simply, I think we should be moving away from supporting users putting in paths that are *actively wrong*, since we now after this PR will give them a really nice error message that ells them how to figure out what paths are valid.

We (ie you guys) could also write a check_path_valid function that could be called at the top of things with pathing support, I suppose, but I'm not sure its worth a full additional walk of the tree, tbh.

DESCRIPTION
gmbecker: Why was this changed @jcheng5, @pawelru @cicdguy ? Are new features actually being used? At the very least it should probably be updated again to 7.3.1 which looks to contain a number of bugfixes of various significance. 

Also, fyi, because 7.3.0 is now not the latest CRAN release and so is only available in source form, AND it requires c++11 features, its pretty non-trivial to install (it currently won't install for me, I'll keep trying but its a lot of wheel spinning thats not really useful to anyone).

gmbecker: Yep, sorry about that other Joe. my mistake there.

I think we should do that update, it seems 7.3.1 is purely bugfixes on top of 7.3.0. 

Also, @cicdguy is there anywhere convenient that the manifest for the images is published?  If not, its probably a good idea since roxygen mismatches cause check failures  the iteration overhead both for the contributor and the nest team is non-trivial. I'm not exactly sure where the best place to put that would be, though, as it would apply across all NEST projects, right? (though I don't know how many contributed PRs the other projects are getting...)

Melkiades (commented):
For me this is good to go. Thank you @gmbecker!

@shajoezhu when you think is best we can merge this PR

shajoezhu (approved):
Thanks Gabe! I think this is good. Let's do one more check in the integration test before we merging this.

edelarua (approved):
No breaking changes in downstream packages. This should be good to go!

shajoezhu:
good to go. thank you all @gmbecker @edelarua @Melkiades

PR #821, opened 2024-01-29
update deps of roxygenize hook

pawelru:


shajoezhu (approved):
lgtm! thanks @pawelru

PR #829, opened 2024-02-22
hotfix for `as_result_df` edge case

Melkiades:
Problem should be also solved at the source of degenerative table: https://github.com/insightsengineering/tern/issues/1195


edelarua (approved):
Lgtm! Thanks!!

PR #819, opened 2024-01-23
785 colsubsetexpr refactor

gmbecker:
Not sure what happened to the original PR. I guess it got destroyed when I rebased on main after doing the inline commit thing (which sadly messed everything up by making things out of sink)

Sorry about but I think the (now lost) comments are addressed here...

PR #856, opened 2024-04-12
support displaying column counts at higher levels in column structure

gmbecker:
clean lint, includes tests, doesn't include updates to vignette(s)


R/argument_conventions.R
```diff
@@ -122,6 +122,11 @@ gen_args <- function(df, alt_counts_df, spl, pos, tt, tr, verbose, colwidths, ob
 #'   uniqueness.
 #' @param vars (`character`)\cr vector of variable names.
 #' @param var_labels (`character`)\cr vector of labels for one or more variables.
+#' @param show_colcounts (`logical(1)`)\cr Should column counts be displayed at the level
```
Melkiades, line 125:
```suggestion
#' @param show_colcounts (`flag`)\cr Should column counts be displayed at the level
```

R/argument_conventions.R
```diff
@@ -122,6 +122,11 @@ gen_args <- function(df, alt_counts_df, spl, pos, tt, tr, verbose, colwidths, ob
...
+#'   facets created by this split. Defaults to `FALSE`.
+#' @param colcount_format (`character(1)`)\cr If `show_colcounts` is `TRUE`, the
```
Melkiades, line 127:
```suggestion
#' @param colcount_format (`string`)\cr If `show_colcounts` is `TRUE`, the
```

R/make_subset_expr.R
```diff
@@ -243,6 +246,9 @@ create_colinfo <- function(lyt, df, rtpos = TreePos(),
       0
     } else {
       vec <- try(eval(ex, envir = alt_counts_df), silent = TRUE)
+      ## likely unneeded now because it happens in splitvec_to_coltree
+      ## which is called during coltree construction above
+      ## TODO remove me
```
Melkiades, line 251: if it is not run at all anymore we can already remove it

R/tt_dotabulation.R
```diff
@@ -1240,7 +1240,7 @@ build_table <- function(lyt, df,
   if (any(alt_params) && is.null(alt_counts_df)) {
     stop(
       "Layout contains afun/cfun functions that have optional parameters ",
-      ".alt_df and/or .alt_df_row, but no alt_count_df was provided in ",
+      ".alt_df and/or .alt_df_row, but no alt_counts_df was provided in ",
```
Melkiades, line 1243: good catch!!

R/tt_toString.R
```diff
@@ -313,6 +314,33 @@ setMethod(
...
 
+
+check_ccount_vis_ok <- function(tt) {
+  ctree <- coltree(tt)
+  tlkids <- tree_children(ctree)
+  lapply(tlkids, ccvis_check_subtree)
+  invisible(NULL)
+}
+
+ccvis_check_subtree <- function(ctree) {
```
Melkiades, line 325: Nice checks!!!

R/tree_accessors.R
```diff
@@ -2307,6 +2364,305 @@ setMethod(
...
+#' facet_colcount(tbl, c("ARM", "A: Drug X", "SEX", "F"))
+#' facet_colcount(tbl, c("ARM", "A: Drug X", "SEX", "F", "STRATA1", "A"))
+#'
+#' ## modify specific count after table creation
+#' facet_colcount(tbl, c("ARM", "A: Drug X", "SEX", "F", "STRATA1", "A")) <- 25
+#'
+#' ## show black space for certain counts by assign NA
+#'
+#' facet_colcount(tbl, c("ARM", "A: Drug X", "SEX", "F", "STRATA1", "C")) <- NA
+setGeneric("facet_colcount",
```
Melkiades, line 2510: facet makes sense but it could be difficult to search for. Maybe `split_colcount` is more intuitive for users that are already using {rtables}

R/tree_accessors.R
```diff
@@ -2307,6 +2364,305 @@ setMethod(
...
+  for (curpath in allpaths[match_paths]) {
+    colcount_visible(obj, curpath) <- value
+  }
+  obj
+}
+
+#' Get or set column count for a facet in column space
+#'
+#' @inheritParams gen_args
+#' @param path character. This path must end on a
```
Melkiades, line 2469:
```suggestion
#' @param path (`string`)\cr this path must end on a
```

R/colby_constructors.R
```diff
@@ -1834,9 +1857,76 @@ manual_cols <- function(..., .lst = list(...)) {
     label = names(.lst)
   ))
   ctree <- splitvec_to_coltree(data.frame(), splvec = splvec, pos = TreePos())
-  InstantiatedColumnInfo(treelyt = ctree)
+  ret <- InstantiatedColumnInfo(treelyt = ctree)
+  rm_all_colcounts(ret)
```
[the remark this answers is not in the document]
gmbecker: Because counts are now calculated during the coltree creation process, which happens even in the case of manually constructed cols, but historically, the column counts for manually constructed columns have defaulted to NA, and in fact there is a test that implicitly enforces that they are missing.

R/tree_accessors.R
gmbecker: splits don't have column counts, because splits are pre-data. once it is instantiated it is a facet. If users don't understand what facet means that is an education failure on our part but split_colcount isn't correct in the technical sense.

R/argument_conventions.R
gmbecker: String also isn't a class in R...

R/argument_conventions.R
gmbecker: flag is not a class in R, it doesn't make sense for it to be treated as such (ie surrouned in ticks)

R/tree_accessors.R
gmbecker: Also I'd think people would be searching for "colcount" to try and figure out how to do this, in which case this would come up

gmbecker:
Also, I don't really have any idea whats going on with the spell checker, it appears to be flagging files that this PR doesn't change (e.g., tt_as_flextable.Rd)

I can't easily use pre-commit in my fork because the hooks have been updated from `0.4.0` to `0.4.2` and it won't even run the checks unless you update the  pre-commit hook yaml file, but then it won't match what is in the insightsengineering repo... @cicdguy any advice on that?

R/tree_accessors.R
gmbecker: This is incorrect. A string, in as much as it exists at all (which, again,  in R it doesnt), has length one, whereas a path is a character vector of any length.

gmbecker: I will admit I forgot to put the weird new stuff around it though...

gmbecker:
> > Also, I don't really have any idea whats going on with the spell checker, it appears to be flagging files that this PR doesn't change (e.g., tt_as_flextable.Rd)
> > I can't easily use pre-commit in my fork because the hooks have been updated from `0.4.0` to `0.4.2` and it won't even run the checks unless you update the  pre-commit hook yaml file, but then it won't match what is in the insightsengineering repo... @cicdguy any advice on that?
> 
> Ok investigatiNg this.

Shot in the dark, if it is running spell check on the full diffs instead of the additions, it could be catching things that were removed. That could explain why it always complains so much and about seemingly random things, but it wouldn't explain complaining about files I didn't touch at all, so I dunno

Melkiades (commented):
@gmbecker @shajoezhu I think before this one we should merge the other outstanding PRs from Gabe. As soon as those are in we can review this PR more in detail. Should we start to get #819 passing the tests?

R/argument_conventions.R
Melkiades: Fine with me. I do not have a strong opinion about this

R/make_subset_expr.R
Melkiades: Should we remove it?

R/tree_accessors.R
```diff
@@ -2307,6 +2364,305 @@ setMethod(
...
+  trimmed_path <- path[c(1, seq(2, length(path), by = 2))]
+
+  do_recursive_replace(obj, trimmed_path, value = value)
+}
+
+
+#' Set visibility of column counts for a group of sibling facets
+#'
+#' @inheritParams gen_args
+#' @param path character. The path *to the parent of the
```
Melkiades, line 2447:
```suggestion
#' @param path (`character`)\cr the path *to the parent of the
```

R/tree_accessors.R
```diff
@@ -2307,6 +2364,305 @@ setMethod(
...
+#' desired siblings*. The last element in the path should
+#' be a split name.
```
Melkiades, line 2449:
```suggestion
#'   desired siblings*. The last element in the path should
#'   be a split name.
```

gmbecker:
Converted to draft because it turns out this requires a large rework how cbinding works befcause the current way of cbinding does not result in column structure where all elements are path-addressable.

I have this working in the combined branch we are working off of at J&J but I haven't cherrypicked the changes to put them into this PR yet. I will do so (or it won't matter if we get the truetype stuff merged in first)

gmbecker:
> @gmbecker @shajoezhu I think before this one we should merge the other outstanding PRs from Gabe. As soon as those are in we can review this PR more in detail. Should we start to get #819 passing the tests?

#819 is subsumed by the truetype font PR, which was derailed by the large scale doc changes causing tons of conflicts. The corresponding formatters PR exists, and I'll work on getting the rtables one up, but that is the one we should use

gmbecker:
Closing this PR and will open another one after #865 is merged in (I have a bundled branch we are doing our work with that has both features so I'll merge that into a new branch after the first PR settles)

PR #867, opened 2024-05-08
Add lifecycle deprecation badges/warnings

### LINKED ISSUE #866: complete implementation of `lifecycle`
https://insightsengineering.github.io/rtables/main/reference/rbind.html
image

- [x] add badge files with `usethis::use_lifecycle()` - please review changes done by this function
  - [x] This will promote `lifecycle` to the Imports and that's ok as we will be using this (see points below). Please don't forget to add this to the `pre-commit.yaml` file as well to keep it in sync.
- [x] for deprecated arguments use `foo(bar = lifecycle::deprecated(), ...) {...}` as per docs: https://lifecycle.r-lib.org/articles/communicate.html#deprecating-an-argument-providing-a-new-default
  - [x] (as per linked docs) use the following condition to check if argument is provided: `if (lifecycle::is_present(arg)) {`. In particular: not `if (!is.null(arg))`.
- [x] use `lifecycle` consistently and replace existing `.Deprecated()` calls: https://github.com/search?q=repo%3Ainsightsengineering%2Frtables%20.Deprecated&type=code
- [x] (supplementary to the above) - add appropriate badges for functions (not only arguments)

edelarua:
Closes #866


edelarua:
I have read the CLA Document and I hereby sign the CLA

Melkiades (approved):
I see you fixed only the deprecated functions. I think we do not need to have the stable badge for now. What you think?

Lgtm anyway ;) Thanks Emily!

edelarua:
> I see you fixed only the deprecated functions. I think we do not need to have the stable badge for now. What you think?

I think the stable badge can be added separately, since it only affects documentation.

Issue #909, opened 2024-08-07
[Bug]: Misalignment between row label and cells


JeanLucMoron:
the export_to_docx functionality works overall very well, however, I have noticed a slight vertical misalignment between the row label and the cell.
Looking at the formating details in MS Word, it seems that while both have vertical alignment = top, there is an additional paragraph setting which differs between both elements: vertical spacing (1pt vs 5 pts).  
![image](https://github.com/user-attachments/assets/99d50360-9e40-4130-b159-ebc7c8f0f5d3)

I have observed this in several standard tables like AET10.

shajoezhu:
hi @JeanLucMoron and @Melkiades , were they using different font? i am guessing it is in the first column, and the font is a bit off too?

JeanLucMoron:
@shajoezhu , they are using the same font (arial ) and size (9)
attaching the word document so that you can see
[aet10 (2).docx](https://github.com/user-attachments/files/16531831/aet10.2.docx)

Issue #903, opened 2024-07-24
Multiple indentation happen for AE tables in the grade columns when exporting to docx


Melkiades:
Issue provided by a user. Will add reproducible code soon

shajoezhu:
A similar issue is reported downstream in autoslider as well regarding indentation

https://github.com/insightsengineering/sme-tasks/issues/721

Melkiades:
related issue: https://github.com/davidgohel/flextable/issues/652

Issue #738, opened 2023-09-25
follow-up to docx export


Melkiades:
Refinements need the following fixes:

- [x] refactor "auto" widths to have `layout = "fixed"` otherwise. See #766 
- [x] topleft vertical alignment and boldness
- [x] topleft could be merged with `flextable::merge_at` but then indentation could be inexact (and row counts need an exception)
- [x] translation of character width to mm or pt (fix width) - Lets rely on Word for wrapping and paginating (closed)
- [ ] #938
- [x] remove somehow `tbl` from theme; independent from `rtables` object (need some thinking for label_row boldness). Maybe can be null and assigned when in `tt_to_flextable` or error?
- [ ] low border in first page is missing if it spans multiple tables (could not find the option to change this behavior)
- [ ] Add a vignette about exports and update README with docx
- [x] merge file template and doc file usage
- [x] cast (...) from export to tt_to_flextable
- [x] add limits 6-12 for font_size
- [x] add selection of different pages (from export not from theme (or both?))
- [x] #924
- [x] #925
- [ ] #926
- [ ] #939

clarkliming:
may I ask if we can allow the table to be exported into a existing file?

Melkiades:
> may I ask if we can allow the table to be exported into a existing file?

You should be able with the template_file option. I did not test it fully though! Meaning that I do not know if the template format will be overloaded or vice-versa. It should just added with the specified style

clarkliming:
I see. I was wondering if we should only use `file` as the only argument, like if `file` exists and is a docx file then append result to it; if `file` does not exist then start from new doc files.

Melkiades:
> I see. I was wondering if we should only use `file` as the only argument, like if `file` exists and is a docx file then append result to it; if `file` does not exist then start from new doc files.

yes it makes sense if you think it is simpler I will add it to this issue

PR #912, opened 2024-08-13
Fix indentation bugs and a couple of other issues

Melkiades:
#903 #909 and some from #738


R/tt_export.R
```diff
@@ -942,7 +964,11 @@ tt_to_flextable <- function(tt,
   flx <- flextable::width(flx, width = final_cwidths) # xxx to fix
 
   if (!is.null(theme)) {
-    flx <- theme(flx)
+    flx <- theme(
+      flx,
+      tbl_ncol_body = NCOL(tt),
+      tbl_row_class = make_row_df(tt)$node_class # These are ignored if not in the theme
+    )
```
Melkiades, line 971: this is why theme should always have ...

shajoezhu: what if we introduce a default, and dummy theme

Melkiades: you mean instead of null?

NEWS.md
```diff
@@ -1,6 +1,9 @@
 ## rtables 0.6.9.9007
 ### Features
  * Modified `reorder_split_levels()` to cover more edge cases and be more stringent in the allowed inputs.
+ * Removed `tt` input from `theme_docx_default()` and added internal handling for row classes and number of columns.
+ * Reworked padding and spacinf in default theme `theme_docx_default()`.
```
Melkiades, line 5:
```suggestion
 * Reworked padding and spacing in default theme `theme_docx_default()`.
```

NEWS.md
```diff
@@ -1,6 +1,9 @@
...
+ * Refactored page size handling for `export_as_docx()`.
```
Melkiades, line 6:
```suggestion
 * Added top left information handling (now bold and bottom aligned)
```

shajoezhu:
shall we try flextable >= 0.9.6 and officer >= 0.6.6, rspm

Melkiades:
> shall we try flextable >= 0.9.6 and officer >= 0.6.6, rspm

I think it may not be too relevant as the only fix that got in that was necessary comes from https://github.com/davidgohel/flextable/issues/652 which would be release 0.6.8 (when it will arrive). Besides, I do not think the different versions are relevant for us so far

shajoezhu (approved):
lgtm! thanks @Melkiades

PR #918, opened 2024-08-21
Update DESCRIPTION

shajoezhu:


Melkiades (approved):
Lgtm! Thanks Joe :)

.github/workflows/scheduled.yaml
```diff
@@ -21,6 +21,8 @@ jobs:
       strategy: ${{ matrix.test-strategy }}
       additional-env-vars: |
         PKG_SYSREQS_DRY_RUN=true
+      extra-deps: |
+        xml2 (>= 1.1.0)
```
pawelru, line 25: this looks redundant

We have the same in DESC file which is enough

shajoezhu, line 25:
```suggestion
```

PR #928, opened 2024-09-03
update green ci

shajoezhu:
close #916

PR #931, opened 2024-09-06
skip before require

pawelru:
follow up on https://github.com/insightsengineering/rtables/pull/928/files

execute skip before any require


shajoezhu (approved):
thanks @pawelru

Issue #925, opened 2024-08-29
Correct colwidths with new font support


Melkiades:

Melkiades:
I did it w/o the font support basing everything on `flextable::dim` with mixed results

PR #937, opened 2024-09-25
Pagination and column widths handling

### LINKED ISSUE #924: Enable page_by pagination for docx

Melkiades:
Fixes #924 and hopefully #925

It seems that the pagination needs a lot of additional work to function correctly. It may be better to work on it in the next PI. @shajoezhu fyi here pagination works but column widths and lpp/cpp estimations are a bit difficult to concretize consistently

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
 ## rtables 0.6.10.9000
 
+### New Features
+ * Experimental pagination is now possible in `tt_as_flextable()` and `export_as_docx()`.
+ * Added handling of widths in `tt_as_flextable()`. Now it is possible to change column widths for `.docx` exports.
+
 ### Bug Fixes
  * Fixed bug that was keeping indentation space characters in top left information when making a `flextable` from a `TableTree` object.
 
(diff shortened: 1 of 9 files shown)


shajoezhu:
i start thinking again that we need another downstream package for exporting

Melkiades:
> i start thinking again that we need another downstream package for exporting

The export functions are in {rtables} when available only for tables and {formatters} when for both tables and listings. Making a separate package would make sense if we wanted to cover more objects, but I do not see this happening soon. Also, exporters are relatively small; the larger function here is the flextable transformer, for which there are a lot of aesthetics that are hard to test. 

In summary, we already have a complex dependency tree, and we might want to keep dev exporters in {rtables} and full exporters (w/ listings capabilities) in {formatters}. Beside that I do not see too much the advantage of having exporters in a separate package atm. Also the risk of circular deps would be very real here, forcing us to move function around beyond what we need to. 

I am open to discussing this anyway! Maybe a pros and cons next PI? :)

Melkiades:
> Fixes #924 and hopefully #925

I managed to make these work, but the estimation of the actual page dimension is very faulty. It will need more design for that

edelarua (approved):
This looks good to go!! It can be revisited at a later date if refinement is needed :)

PR #978, opened 2024-12-18
Replace logo.png by logo.svg

llrs-roche:
# Pull Request

Replace the logo.png by logo.svg to reduce the package size (500Kb to 156Kb for the logo).

Apparently the svg is not picked up by roxygen2 to be shown on rtables-package.Rd.

Fixes https://github.com/insightsengineering/nestdevs-tasks/issues/93

PR #979, opened 2024-12-19
Rename logo to be displayed

llrs-roche:
Apologies, I left the change uncommitted when I pushed on #978 with this PR the logo should display correctly

image


Melkiades (approved):
Thanks @llrs-roche!

Issue #976, opened 2024-12-17
Consider making the table names as an optional addition of `make_ard = TRUE`


Melkiades:

shajoezhu:
isn't this done already?

shajoezhu:
ok, not done yet, but i agree, would be nice

Issue #970, opened 2024-12-12
[Bug]: When using split_cols_by_multivar in layout, make_ard=T doesn't work


vchang-cims:
When building a table with `split_cols_by_multivar` in the layout, and then passing the table into `as_result_df(make_ard = TRUE)`, the following error occurs:
```
'Error in data.frame(flattened_cols_names) : 
  row names contain missing values'
```

Here is a reproducible example:
```
n <- 400

df <- tibble(
  arm = factor(sample(c("Arm A", "Arm B"), n, replace = TRUE), levels = c("Arm A", "Arm B")),
  country = factor(sample(c("CAN", "USA"), n, replace = TRUE, prob = c(.55, .45)), levels = c("CAN", "USA")),
  gender = factor(sample(c("Female", "Male"), n, replace = TRUE), levels = c("Female", "Male")),
  handed = factor(sample(c("Left", "Right"), n, prob = c(.6, .4), replace = TRUE), levels = c("Left", "Right")),
  age = rchisq(n, 30) + 10
) %>% mutate(
  weight = 35 * rnorm(n, sd = .5) + ifelse(gender == "Female", 140, 180)
)

colfuns <- list(
  function(x) in_rows(mean = mean(x), .formats = "xx.x"),
  function(x) in_rows("# x > 5" = sum(x > .5), .formats = "xx")
)


lyt <- basic_table() %>%
  split_cols_by("arm") %>%
  split_cols_by_multivar(c("age", "weight")) %>%
  split_rows_by("country"
  ) %>%
  summarize_row_groups() %>%
  analyze_colvars(afun = colfuns)


tbl <- build_table(lyt, df)
tbl
as_result_df(tbl,make_ard = T)
```
Thank you!

shajoezhu:
hi @vchang-cims , thanks a lot for raising this. we are actively working on this. thanks for bring this to our attention

PR #986, opened 2025-01-27
add precision column `stat_string` and different table names

### LINKED ISSUE #977: Add stat_string for current print precision number

Melkiades:
Fix #977 #976 #970


edelarua (approved):
This looks good!! Thanks Davide! Just one typo in the NEWS file to correct.

PR #996, opened 2025-02-12
fix every row sep behavior of analyze section_div

gmbecker:
This is technically a breaking change, but one we all agree on as far as I understand.

`section_div` in `analyze` will now always apply to analyze blocks and never apply to each row generated by the analyze

tests are updated to reflect the new behavior


Melkiades (commented):
@gmbecker Thank you for your PR. Could you add a way to get the previous functionality in too? I would add another parameter to analyze function to do it. I would prefer not adding it to `in_rows` for giving current users an easy fix

tests/testthat/test-printing.R
```diff
@@ -758,12 +788,10 @@ test_that("Separators and wrapping work together with getter and setters", {
     "thing its          ",
     "so                 ",
     "  m            8   ",
-    "                   ",
```
shajoezhu, line 761: i suspect this going to lead a few breaking changes downstream

Melkiades:
> @gmbecker Thank you for your PR. Could you add a way to get the previous functionality in too? I would add another parameter to analyze function to do it. I would prefer not adding it to `in_rows` for giving current users an easy fix

Thinking about it more, I think this should be an option in in_rows to set the trailing_sep and not in analyze()

tests/testthat/test-printing.R
gmbecker: It could be, but as we discussed, theres no way to fix an inconsistency without a breaking change, and fixing it in the other direction will, i suspect, break *much more* downstream code on both your side and ours

NEWS.md
```diff
@@ -1,3 +1,4 @@
+
```
Melkiades, line 1:
```suggestion
```

Melkiades (approved):
Lgtm! Thanks Gabe. I would add the possibility to add a value in `in_rows` but it is not too necessary atm. Also tested getter and setter and it is all fine for me

shajoezhu (approved):
Thanks Gabe! Downstream all checks out

https://github.com/insightsengineering/scda.test/pull/185

I was wondering if you could add a row_div option as we discussed last time, we can follow-up when meet next week

Melkiades:
> Thanks Gabe! Downstream all checks out
> 
> [insightsengineering/scda.test#185](https://github.com/insightsengineering/scda.test/pull/185)
> 
> I was wondering if you could add a row_div option as we discussed last time, we can follow-up when meet next week

That would be great!! Could you keep how the setter works now? ;) (you can add line by line trailing_sep atm). Thanks

PR #1015, opened 2025-03-31
add ability to pass arguments to score and prune functions

gmbecker:


Melkiades (commented):
It looks good to me. Thanks Gabe! Could you add a news entry?

Melkiades (approved):
Good to go! Thanks Gabe

Issue #1010, opened 2025-03-19
Breaking change between 0.6.8 and 0.6.11


gmbecker:
There appears to have been a regression/breaking change sometime between 0.6.8 and 0.6.11. The issue occurs when splitting on character variables where not all values appear in all parent facets.  Consider the following code:

```r
library(dplyr)
library(rtables)
library(tern)
library(stringr)

### manually run below code
packageVersion("rtables")


adsl <- random.cdisc.data::cadsl
adeg <- random.cdisc.data::cadeg

################################################################################
# Define script level parameters:
################################################################################
adsl_f <- adsl %>%
  select(USUBJID, ACTARM)

# Note:  We keep only post-baseline for analysis.
adeg_f <- adeg %>%
  filter(AVISIT %in% levels(adeg$AVISIT)[1:3]) %>%
  filter(PARAM %in% c("Heart Rate", "QT Duration", "RR Duration")) %>%
  var_relabel(
    PARAM = "Assessment",
    ANRIND = "Abnormality"
  ) %>% 
  select(USUBJID, AVISIT, PARAM, PARAMCD, ANRIND, ACTARM) %>% 
  mutate(PARCAT = case_when(PARAMCD == "QT" ~ "QT",
                            TRUE ~ "Other")) %>% 
  mutate(ANRIND2 = paste(ANRIND, PARAMCD))  %>% 
  mutate(ANRIND3 = paste(ANRIND, PARAM))

# Ensure character variables are converted to factors and empty strings and NAs are explicit missing levels.
adsl_f <- df_explicit_na(adsl_f)
adeg_f <- df_explicit_na(adeg_f)

adeg_f_1 <- adeg_f %>% 
  filter(PARAMCD != "QT") %>% 
  mutate(CRITDIR = "Direction 1") %>% 
  mutate(ANRIND23 = ANRIND2)  

adeg_f_2 <- adeg_f %>% 
  mutate(CRITDIR = "Direction 2") %>% 
  mutate(ANRIND23 = ANRIND3)

adeg_f <-rbind(adeg_f_1, adeg_f_2)



### mapping dataframe 

mapdf <- unique(adeg_f %>% select(PARCAT, PARAMCD, CRITDIR, ANRIND23)) %>% 
  arrange(PARCAT, PARAMCD, CRITDIR, ANRIND23) %>% 
  mutate(across(where(is.factor), as.character))

### 3 versions of dataset
### CRITDIR not factor, all visits have available data for all PARAMCD
[...]

PR #1017, opened 2025-04-03
`keep_split_levels()` correction for empty branches and more descriptive error

Melkiades:
#1010

### DIFF
--- a/R/default_split_funs.R
+++ b/R/default_split_funs.R
@@ -73,9 +73,10 @@ keep_split_levels <- function(only, reorder = TRUE) {
     unique_vals <- .get_unique_levels(varvec)
 
     # Error in case not all levels are present
-    if (!all(only %in% unique_vals)) {
+    if (length(varvec) > 0 && !all(only %in% unique_vals)) {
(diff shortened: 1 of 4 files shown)


R/default_split_funs.R
```diff
@@ -73,9 +73,10 @@ keep_split_levels <- function(only, reorder = TRUE) {
     unique_vals <- .get_unique_levels(varvec)
 
     # Error in case not all levels are present
-    if (!all(only %in% unique_vals)) {
+    if (length(varvec) > 0 && !all(only %in% unique_vals)) {
```
Melkiades, line 76: If we want only to check for factors we can readd `is.factor(varvec)` instead of the above change. I much prefer to check also if the vector is a character if the selected values are present.

edelarua: I agree we should account for character vectors as well.

edelarua (approved):
Lgtm!! Thanks Davide

R/default_split_funs.R
Melkiades: I think it is settled then! Thanks Emily :)

PR #1026, opened 2025-06-02
export subset_cols and has_force_pag

### LINKED ISSUE #1025: export a couple of functions for use in packages like junco
In junco (to be open sourced shortly) we needed to use 

```
rtables:::has_force_pag
rtables:::subset_cols
```

Can we please export these two functions, even if they are documented as not intended for end-user use so that packages can use them?

gmbecker:
clean lint for me locally. NEWS updated. Explicit tests as requested.

Closes #1025


R/tt_pos_and_access.R
```diff
@@ -843,7 +843,33 @@ setMethod(
 
 ## we have two options here: path like we do with rows and positional
 ## in leaf space.
-
+#' Subset a table or row to particular columns
+#' @inheritParams gen_args
+#' @inheritParams brackets
+#' @inheritParams head
+#' @param j (`integer`, `logical` or `character`)\cr The column(s) to subset `tt`
+#'   down to. Character vectors are intepreted as a *column path*, not as names.
```
Melkiades, line 851:
```suggestion
#'   down to. Character vectors are interpreted as a *column path*, not as names.
```

PR #1038, opened 2025-10-07
Fix for upcoming testthat 3.3.0 release

hadley:
`expect()` now checks its inputs highlighting that you accidentally used it instead of the correct expectation


tests/testthat/test-lyt-tabulation.R
```diff
@@ -10,8 +10,6 @@ test_that("summarize_row_groups works with provided funcs", {
 
   tb1 <- build_table(l1, DM)
```
Melkiades, line 11:
```suggestion
  expect_silent(
    tb1 <- build_table(l1, DM)
  )
```

PR #1044, opened 2025-10-23
vignette on translating shells to layouts

gmbecker:


gmbecker:
Closes #1043

gmbecker:
@shajoezhu  all checks now passing except spell check which i intentionally skipped because it continues to give me trouble with accented letters such as those in Daniels name.

PR #1063, opened 2025-12-09
fix candidate for 1058

gmbecker:
downstream tests: https://github.com/insightsengineering/scda.test/pull/202


shajoezhu (approved):
lgtm! thanks @gmbecker

PR #1081, opened 2026-03-23
add restrict_facets and export make_subset_expr

gmbecker:


Melkiades (commented):
I think it is good to go! Thanks for the additions. I would only add a couple of examples to the documentation

gmbecker:
added some basic examples as requested

Melkiades (approved):
For me is good to go! Thanks Gabe :)

PR #1091, opened 2026-06-03
Add guided advanced vignettes

gmbecker:
and needed fixes/utilities for them


DESCRIPTION
```diff
@@ -105,3 +106,4 @@ Collate:
     'tt_from_df.R'
     'validate_table_struct.R'
     'zzz_constants.R'
+Config/roxygen2/version: 8.0.0
```
Melkiades: Repeated entry

DESCRIPTION
```diff
@@ -1,5 +1,6 @@
 Package: rtables
 Title: Reporting Tables
+
```
Melkiades:
```suggestion
```

R/tree_accessors.R
```diff
@@ -1199,6 +1221,10 @@ setMethod("obj_na_str<-", "Split", function(obj, value) {
 #' @export
 setMethod("obj_na_str", "VTableNodeInfo", function(obj) obj@na_str)
 
+#' @rdname int_methods
+#' @export
+setMethod("obj_na_str", "VTableNodeInfo", function(obj) attr(obj, "row_na_strs", exact = TRUE))
```
Melkiades: Duplicated with the above. I guess it should be:
`setMethod("obj_na_str", "RowsVerticalSection", function(obj) attr(obj, "row_na_strs", exact = TRUE))
`

R/tt_pos_and_access.R
```diff
@@ -1584,6 +1584,14 @@ setMethod(
   }
 )
 
+#' @rdname int_methods
+#' @keywords internal
+#' @exportMethod cell_values
+setMethod(
+  "cell_values", "RowsVerticalSection",
+  function(tt, rowpath, colpath = NULL, omit_labrows = TRUE) {}
```
Melkiades: This would return NULL. why should it? Could you add a comment and maybe a more explicit NULL value?

vignettes/guided_advanced_afuns_rowsverticalsection.Rmd
```diff
@@ -0,0 +1,261 @@
...
+
+# Appendix A - Code For `c.RowsVerticalSection`
+
+The following is a copy of a development veresion of `c.RowsVerticalSection` at the time of writing this vignette; this code is not guaranteed to be kept in sync with `rtables`' exported version of `c.RowsVerticalSection` and should be used for illustrative and back-porting purposes only.
+
+```{r, eval = FALSE}
+c.RowsVerticalSection <- function(...) {
+  lst <- list(...)
+  if (!all(sapply(lst, function(x) inherits(x, "RowsVerticalSection")))) {
+    stop("Cannot use c() to combine RowsVerticalSection objects with objects of other clases")
```
Melkiades:
```suggestion
    stop("Cannot use c() to combine RowsVerticalSection objects with objects of other classes")
```

R/00tabletrees.R
```diff
@@ -2182,6 +2182,43 @@ print.RowsVerticalSection <- function(x, ...) {
...
+#' by [in_rows()]) into a single object
+#'
+#' @param ... RowsVerticalSection objects
+#' @returns A single RowsVerticalSection object containing all
+#'   row sections from the objects passed to `...`
+#' @export
+c.RowsVerticalSection <- function(...) {
+  lst <- list(...)
+  if (!all(sapply(lst, function(x) inherits(x, "RowsVerticalSection")))) {
+    stop("Cannot use c() to combine RowsVerticalSection objects with objects of other clases")
```
Melkiades:
```suggestion
    stop("Cannot use c() to combine RowsVerticalSection objects with objects of other classes")
```

R/00tabletrees.R
```diff
@@ -2182,6 +2182,43 @@ print.RowsVerticalSection <- function(x, ...) {
...
+#' Combine two or more RowsVerticalSection objects (as returned
+#' by [in_rows()]) into a single object
+#'
+#' @param ... RowsVerticalSection objects
+#' @returns A single RowsVerticalSection object containing all
+#'   row sections from the objects passed to `...`
+#' @export
+c.RowsVerticalSection <- function(...) {
+  lst <- list(...)
+  if (!all(sapply(lst, function(x) inherits(x, "RowsVerticalSection")))) {
```
Melkiades, line 2195: I would use `vapply(lst, inherits, logical(1), "RowsVerticalSection")` as it is a expected logical

inst/WORDLIST
```diff
@@ -1,66 +1,110 @@
...
 AE
 AEs
 afun
 afuns
+allcols
 amongst
 ARD
 ard
 ARDs
+args
```
Melkiades, line 12: 68 new words is a lot. Some look like they could be R object names or code fragments that should be in backticks in the vignettes rather than whitelisted (e.g., afun, afuns, bbbs, splfun). Worth checking if wrapping them in backticks in the Rmd files removes the need for the wordlist entries

R/make_split_fun.R
```diff
@@ -394,7 +394,7 @@ add_overall_facet <- function(name, label, extra = list()) {
 #'
 #' @param innervar (`character`)\cr the variable(s) to trim (remove unobserved levels) independently within each facet.
 #'
-#' @return A function suitable for use in the `pre` (list) argument of `make_split_fun`.
+#' @return A function suitable for use in the `post` (list) argument of `make_split_fun`.
```
Melkiades, line 397: nice catch!

R/tree_accessors.R
```diff
@@ -1678,7 +1704,7 @@ setMethod(
         "must have length 1 or the number of rows"
       )
     }
-    attr(obj, "indent_mods") <- as.integer(value)
+    attr(obj, "indent_mods") <- rep(as.integer(value), length.out = length(obj))
```
Melkiades, line 1707: This is probably the right behavior, but it should be documented in NEWS since it changes the semantics of the setter

Melkiades (commented):
Thanks Gabe for the PR!! I added a couple of comments ^^ 

In general, I would suggest to add link to the issue and add the NEWS file when ready to review. If possible I would do bug fixes and docs improvement in different PRs, but it is not an issue

gmbecker:
Sorry @Melkiades this probably should have been marked as a draft, but thanks for going through it and I will go through all your comments and respond to each (and likely make most suggested changes)

inst/WORDLIST
gmbecker: will  look into this,s ome of those are real (bbbs stands for Behavioral Building Blocks, which is a term I use heavily in one of the vignettes to refer to the functions we pass to `pre` and `post` in make_split_fun, others do look like variable names though

R/tree_accessors.R
gmbecker: correct nice catch. thanks

R/tree_accessors.R
gmbecker: this is a bugfix rather than a semantics change because if you give it a single value something down stream (printing, I think) breaks, but still worth putting in the NEWS file

R/tt_pos_and_access.R
gmbecker: This is missing code not intended to return NULL so I'll either remove it (if its not needed) or fix it.

Nice catch thanks.

DESCRIPTION
gmbecker: Thats weird, as it wasn't manually added, but I'll look into this

gmbecker:
superceded by #1093 I believe all of the review comments here have been addressed with the current state there.

PR #1099, opened 2026-07-01
Clean up (re-document, fix linters, update old magritr to native pipe)

munoztd0:
As per we discussed runned a little clean up to be able to do the next PRs with ease of mind.


R/package.R
```diff
@@ -1,10 +1,2 @@
 #' @keywords internal
 "_PACKAGE"
-
-#' @importFrom utils head head.matrix tail tail.matrix
-#' @importFrom stats setNames na.omit prop.test binom.test relevel quantile
-#' @importFrom htmltools tags tagList
-#' @importFrom lifecycle deprecated
-#' @importFrom magrittr %>%
-#' @import methods
```
Melkiades, line 9: this breaks R CMD check, `is()` and `new()` are not found (3 WARN 5 NOTE). rtables is all S4 so we need methods. keep `@import methods` and re-add `importFrom(stats, setNames, na.omit, quantile)` + `importFrom(utils, head, tail, head.matrix, tail.matrix)`. the rest (prop.test/binom.test/relevel/tagList/%>%) are unused now so fine to drop

Melkiades (changes requested):
nice cleanup, pipe migration is thorough. one blocker though:

removing `import(methods)` (and the stats/utils importFroms) breaks the namespace load. R CMD check is 3 WARN 5 NOTE with `could not find function "is"` / `"new"`. load_all hides it because it attaches the deps, only a real install/check shows it.

re-add `@import methods` + `importFrom(stats, setNames, na.omit, quantile)` + `importFrom(utils, head, tail, head.matrix, tail.matrix)` and it's good. prop.test/binom.test/relevel/tagList/%>% are actually unused now so dropping those is fine.

shajoezhu:
omg! This finally happens! thanks so much man! @munoztd0

R/package.R
munoztd0: wow totally my bad, I just wanted to remove magrittr and I must have been a bit to "clicky"

munoztd0:
> omg! This finally happens! thanks so much man! @munoztd0

Anytime! Please let me know if you need this kinda of help enywhere else in the NEST ecosytem, I am always here to help/support it.

Melkiades (approved):
Good to go!! Thanks @munoztd0

Issue #1089, opened 2026-05-12
Feature request: passing .alt_df_full to afuns when using build_table(lyt, df)


iaugusty:
Would it be possible to consider passing the input dataframe df to .alt_df_full when build_table has been called with no alt_counts_df dataframe?

The current behaviour requires to set a default for .alt_df_full in the afun as demonstrated in following reprex. 

``` r
library(rtables)
#> Loading required package: formatters
#> 
#> Attaching package: 'formatters'
#> The following object is masked from 'package:base':
#> 
#>     %||%
#> Loading required package: magrittr
#> 
#> Attaching package: 'rtables'
#> The following object is masked from 'package:utils':
#> 
#>     str
adsl <- ex_adsl

myafun <- function(df, .var, .alt_df_full, .spl_context){
  xx <- .alt_df_full
  yy <- .spl_context$full_parent_df[[1]]
  in_rows(.list = list(full_parent_df = NROW(yy), 
                       .alt_df_full = NROW(xx),
                       df = NROW(df)),
          .labels = c("# (full_parent_df)",
                      "# (.alt_df_full)",
                      "# (df)"))
}

lyt <- basic_table() |> 
  split_cols_by("ARM") |> 
  analyze(vars = "STUDYID", afun = myafun)

build_table(lyt, adsl, alt_counts_df = adsl)
#>                      A: Drug X   B: Placebo   C: Combination
#> ————————————————————————————————————————————————————————————
#> # (full_parent_df)      400         400            400      
#> # (.alt_df_full)        400         400            400      
#> # (df)                  134         134            132
build_table(lyt, adsl)
#> Error: Error applying analysis function (var - STUDYID): argument ".alt_df_full" is missing, with no default
#>  occured at (row) path: root

# set default for .alt_df_full to NULL when not passed by rtables splitting machinery

myafun2 <- function(df, .var, .alt_df_full = NULL, .spl_context){
  xx <- .alt_df_full
  yy <- .spl_context$full_parent_df[[1]]
  in_rows(.list = list(full_parent_df = NROW(yy), 
                       .alt_df_full = NROW(xx),
                       df = NROW(df)),
          .labels = c("# (full_parent_df)",
                      "# (.alt_df_full)",
                      "# (df)"))
}

lyt2 <- basic_table() |> 
  split_cols_by("ARM") |> 
  analyze(vars = "STUDYID", afun = myafun2)

build_table(lyt2, adsl)
[...]

PR #1102, opened 2026-07-31
1089 always avail alt df

gmbecker:


gmbecker:
fixes #1089

R/colby_constructors.R
```diff
@@ -1044,7 +1044,11 @@ NULL
 #'   \item{.all_col_exprs}{List of expressions. Each of them represents a different column splitting.}
 #'   \item{.all_col_counts}{Vector of integers. Each of them represents the global count for each column. It differs
 #'     if `alt_counts_df` is used (see [build_table()]).}
-#' }
+#'
+#' For the `.alt_df*` family of parameters, these will be passed data
+#' subsets based on `df` if no `alt_counts_df` is specified in the
+#' `build_table` call. In `rtables` versions `<= 0.6.13`, this instead
+#' resulted in an error.  }
```
Melkiades, line 1051: @gmbecker small thing on the version here,I checked the tags and the `stop()` was actually still in `v0.6.16`, it is only being removed in this PR. So `<= 0.6.13` is not right, it should point to the last release before this fix (`0.6.16`) or just say "in previous versions" to be safe. Also there is a double space and the closing `}` got glued onto the sentence (`error.  }`), and this paragraph is sitting inside the `\describe{}` block, so it will render as a weird list item. I would move it out just after the closing brace so it reads as body text.

R/colby_constructors.R
```diff
@@ -1044,7 +1044,11 @@ NULL
...
 #'
 #' @note If any of these formals is specified incorrectly or not present in the tabulation machinery, it will be
```
Melkiades, line 1053: Related to the note just below (line 1056): it still says `.alt_df_row` and `.alt_df` "will not be present" when no `alt_counts_df` is given, but that is exactly what we are changing here, so it now contradicts the paragraph right above. Can we update/drop it? Otherwise it is a bit confusing for the user ;)

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -384,3 +377,31 @@ test_that("full alt_counts_df is accessible from afun/cfun via .alt_df_full", {
   cvals <- unlist(cell_values(tbl))
   expect_true(all(cvals == "ok"))
 })
+
+test_that(".alt_df* argument behavior is correct when alt_counts_df is not set", {
+  check_alt_dfs <- function(df, .df_row, .alt_df_row, .alt_df, .alt_df_full) {
+    expect_identical(df, .alt_df)
+    expect_identical(.df_row, .alt_df_row)
+    expect_false(is.null(.alt_df_full))
```
Melkiades, line 385: Nice test. One thing, here we only check `.alt_df_full` is not `NULL`, but the whole point is that it is the *full* `df` (not the row-group subset). I would add an `expect_identical(.alt_df_full, ex_adsl)` like the existing "full alt_counts_df is accessible" test does, otherwise a regression where it collapses to the subset would still pass here.

tests/testthat/test-tab_afun_cfun.R
```diff
@@ -166,9 +166,6 @@ test_that(".spl_context and afun extra parameters contain information about comb
 
   # NB: If you add keep_levels = c("all_X") to add_combo_levels the other
   #     column expressions are missing -> Expected!
-  expect_error(lyt |> build_table(DM),
-    regexp = "Layout contains afun\\/cfun functions that have optional*"
-  )
```
Melkiades, line 169: Minor leftover: removing the old `expect_error` left a dangling blank line here, the `keep_levels` comment now floats away from the `build_table` call it refers to. Just yo tidy up.

Melkiades:
any news on this @gmbecker?

Issue #1095, opened 2026-06-12
Allow Intermediate Levels Of Nesting


gmbecker:
Instead of just `TRUE`, `FALSE`, ***if possible***

PR #1111, opened 2026-08-17
1095 intermediate nesting

gmbecker:
Intermediate nesting behavior, testing and documentation


R/nesting_impl.R
```diff
@@ -188,12 +193,12 @@ get_names_list <- function(splvec) {
     if (is(x, "SplitVectorTree")) {
       c(
         ## use this cause it does deuniqify
-        list(vapply(x, rtables:::first_spl_name, "")),
+        list(vapply(x, first_spl_name, "")),
         ## ignore first name of last branch, we use name from first branch for matching here
         get_names_list(x[[length(x)]][-1])
       )
     } else { ## Split case
-      rtables:::first_spl_name(x)
```
Melkiades: We should not have these in. could you double check there are not any other istances of :::? Thanks!

munoztd0: this has already been resolved

munoztd0 (approved):
finished last tweaks and adress comments

Melkiades (commented):
Nice work, the at_sibling feature does what #1095 asked and the happy path is solid. Verified it builds correctly and the moved column-nesting test still checks out. A couple of real bugs on the error paths though, plus some cleanup. Comments inline.

R/nesting_impl.R
```diff
@@ -0,0 +1,552 @@
...
+  }
+  lastel <- splvec[[branch_pos]]
+
+  len <- length(splvec)
+
+  endontree <- is(lastel, "SplitVectorTree")
+  sib_matches <- is.null(at_sibling) || at_sibling == first_spl_name(lastel)
+  if (endontree && sib_matches) {
+    splvec[[branch_pos]] <- SplitVectorTree(lst = c(lastel, list(SplitVector(newspl))))
+  } else if (has_force_pag(lastel)) {
```
Melkiades, line 272: This crashes on the exact case the docs say is unsupported. If at_sibling targets something that was itself placed with at_sibling, lastel is a SplitVectorTree and has_force_pag has no method for that class, so you get an internal dispatch error instead of the informative message.

```r
basic_table() |>
  split_rows_by("STRATA1") |>
  split_rows_by("SEX") |> analyze("AGE") |>
  split_rows_by("RACE", at_sibling = "SEX") |> analyze("AGE") |>
  split_rows_by("BMRKR2", at_sibling = "RACE") |> analyze("AGE") |>
  build_table(ex_adsl)
#> unable to find an inherited method for 'has_force_pag' for signature 'SplitVectorTree'
```

Needs an up-front guard with the documented error, or a has_force_pag method for SplitVectorTree. Worth a test too.

R/nesting_impl.R
```diff
@@ -0,0 +1,552 @@
...
+    stop(
+      "at_sibling pointed to a split with forced pagination (page_by = TRUE).",
```
Melkiades, line 274: This guard isn't actually reached when the target uses page_by = TRUE. Instead of this message you get 'Got a page title prefix for an Elementary Table':

```r
basic_table() |> split_rows_by("STRATA1", page_by = TRUE) |>
  split_rows_by("RACE", at_sibling = "STRATA1") |> analyze("AGE") |>
  build_table(ex_adsl)
```

Would be good to make the intended error fire, with a test.

R/tt_dotabulation.R
```diff
@@ -882,6 +882,42 @@ setMethod(
...
+    ret <- lapply(
+      spl,
+      function(splvecii) {
+        recursive_applysplit(
+          df = df,
+          lvl = lvl + 1L,
+          alt_df = alt_df,
+          alt_df_full = alt_df_full,
+          splvec = splvecii,
+          name = obj_name(unlist(splvecii, recursive = TRUE)[[1]]), ## XXX I think this is wrong
```
Melkiades, line 908: These XXX markers are in the tabulation core (name= here, no_outer_tbl = TRUE just below and at 1245). Since this drives the actual table build, can we resolve them or pin the assumption with a test rather than shipping the uncertainty?

R/argument_conventions.R
```diff
@@ -150,6 +150,9 @@ gen_args <- function(df, alt_counts_df, spl, pos, tt, tr, verbose, colwidths, ob
 #'   to the *split* or *group of sibling analyses*, for `split_rows_by*` and
 #' `analyze*` when analyzing more than one variable, respectively. Ignored when
 #' analyzing a single variable.
+#' @param at_sibling (`charactere(1)` or `NULL`)\cr If non-null, a preceding
```
Melkiades: typo: charactere(1) -> character(1). This leaks into the generated Rd files.

R/nesting_impl.R
```diff
@@ -0,0 +1,552 @@
...
+    }
+    lyt
+  }
+)
+
+#' @rdname int_methods
+setMethod(
+  "split_rows", "ANY",
+  function(lyt, spl, pos, at_sibling = NULL) {
+    stop("nope. can't add a row split to that (", class(lyt), "). contact the maintaner.")
```
Melkiades: typo in the user-facing message: maintaner -> maintainer (also lines 326, 479, 549, and AnalzyeMultiVars in the comment at 453). Spell-check CI may catch these.

R/nesting_impl.R
```diff
@@ -0,0 +1,552 @@
...
+  nmlst <- get_names_list(splvec)
+
+  atsib <- deuniqify_path_elements(at_sibling)
+  dup_pos <- extract_dup_pos(at_sibling)
+  found_lgl <- vapply(nmlst, function(lst) atsib %in% deuniqify_path_elements(lst), FALSE)
+  found <- which(found_lgl)
+
+  ## i <- 1
+  ## found <- numeric()
+  ## while (i <= length(nmlst) && !found) {
```
Melkiades: Leftover commented-out while loop from the old implementation. Can drop it now that the vapply version above replaces it.

R/nesting_impl.R
munoztd0: Change so that now it fires:

```
basic_table() |> split_rows_by("STRATA1", page_by = TRUE) |>
      split_rows_by("RACE", at_sibling = "STRATA1") |> analyze("AGE
    ") |>
      build_table(ex_adsl)
Error in h(simpleError(msg, call)) : 
  error in evaluating the argument 'obj' in selecting a method for function 'last_rowsplit': at_sibling pointed to a split with forced pagination (page_by = TRUE). This is not supported.
  
  ````

File: R/nesting_impl.R (hunk 6 above)
Comment (gmbecker) [reply R/nesting_impl.R]: The problem with Spellcheck- CI is that it *always* flags Daniel SB's name when I run it locally which causes huge churn in WORDLIST that then needs to be undone in the github action :(

File: R/nesting_impl.R (hunk 2 above)
Comment (gmbecker) [reply R/nesting_impl.R:334]: This is now supported, because it was equally hard to get a properly informative error message as it was to just support it, even though conceptually it doesn't make a ton of sense

File: R/tt_dotabulation.R (hunk 4 above)
Comment (munoztd0) [reply R/tt_dotabulation.R:908]: @gmbecker ?

Comment (gmbecker) [reply R/tt_dotabulation.R:908]: I will look at these and resolve them

Comment (gmbecker) [reply R/tt_dotabulation.R]: I'm confident the behavior is correct so I've removed these

**Comment by gmbecker** [comment]
@munoztd0  please have a read through the nesting vignette, which is now complete

File: R/nesting_impl.R [hunk 8]
```diff
@@ -0,0 +1,621 @@
...
+  function(lyt, spl, constructor) {
+    pos <- length(lyt)
+    tmp <- cmpnd_last_rowsplit(lyt[[pos]], spl, constructor)
+    lyt[[pos]] <- tmp
+    lyt
+  }
+)
+#' @rdname int_methods
+setMethod(
+  "cmpnd_last_rowsplit", "SplitVector",
```
Melkiades, line 507: This path does not know about `SplitVectorTree`, and the result is silent.

```r
lyt <- basic_table() |>
  split_rows_by("STRATA1") |> split_rows_by("SEX") |> analyze("AGE", table_names = "a1") |>
  split_rows_by("RACE", at_sibling = "SEX") |> analyze("AGE", table_names = "a2") |>
  analyze("BMRKR1")
build_table(lyt, ex_adsl)
```

`SEX` and `RACE` both come out childless, row paths stop at `STRATA1/A/SEX`, and the AGE mean under A is 34.19, which is the whole `STRATA1 == A` mean rather than 32.38 for `SEX == F`. Drop the last `analyze()` and it is correct. No error or warning, only the unrelated table-name message.

`analyze()` takes the compounding branch when `nested && is.null(at_sibling) && is_analyze_spl`, and `lst` here is the `SplitVectorTree`, so it lands as `AnalyzeMultiVars(.payload = list(<SplitVectorTree>, spl))`.

I narrowed the trigger: only a bare `analyze()` after `analyze()` following a branch. All of these are fine, a single analyze, `analyze(c("AGE", "BMRKR1"))` in one call, analyze then split then analyze, `summarize_row_groups()` then analyze, and a second analyze carrying its own `at_sibling`.

R/00tabletrees.R
```diff
@@ -1936,7 +1939,28 @@ SplitVector <- function(x = NULL,
   new("SplitVector", lst)
 }
 
+setClass("SplitVectorTree",
```
Melkiades, line 1942: `vars_in_layout()` is exported and errors on anything built with `at_sibling`:

```r
vars_in_layout(lyt)
#> unable to find an inherited method for function 'vars_in_layout' for signature 'lyt = "SplitVectorTree"'
```

I swept every generic carrying a `SplitVector` method against the tree methods added here. The ones without a counterpart are `vars_in_layout`, `cmpnd_last_rowsplit`, `show`, `.tl_indent_inner` and `c`. `cmpnd_last_rowsplit` is the blocker above and `show` is the print comment below, so `vars_in_layout` and `.tl_indent_inner` are the remainder. `c` looks fine in practice since the class contains `list`.

Might be worth walking that list once deliberately rather than adding tree methods as they surface.

R/nesting_impl.R
```diff
@@ -0,0 +1,621 @@
...
+  "get_anchor_list", "PreDataTableLayouts",
+  function(splvec) {
+    get_anchor_list(rlayout(splvec))
+  }
+)
+
+#' @rdname get_anchor_list
+#' @export
+setMethod(
+  "get_anchor_list", "PreDataRowLayout",
```
Melkiades, line 229: This maps over every top-level chain, but resolution only ever looks at `lyt[[pos]]`, the active one. So the accessor and the validator disagree:

```r
lyt <- basic_table() |>
  split_rows_by("SEX") |> analyze("AGE") |>
  split_rows_by("BMRKR2", nested = FALSE) |> split_rows_by("RACE") |> analyze("AGE")

get_anchor_list(lyt)
#> [1] "SEX" "BMRKR2" "RACE" "AGE"

lyt |> split_rows_by("STRATA1", at_sibling = "SEX")
#> Error: Unable to find structural element 'SEX' to add siblings for.
#>   Eligible elements: 'BMRKR2', 'RACE', 'AGE'
```

Since this is the documented way to discover anchors, the two should agree. The error message already computes the right set.

Smaller and separate: the return type moves with the input, `character` when there is no branch, nested `list` once there is one. The roxygen does say "possibly nested list" so it is disclosed, but it makes the result awkward to consume.

R/nesting_impl.R
```diff
@@ -0,0 +1,621 @@
...
+)
+
+#' @rdname get_anchor_list
+#' @export
+setMethod(
+  "get_anchor_list", "Split",
+  function(splvec) first_spl_name(splvec)
+)
+
+find_branch_pos2 <- function(splvec, at_sibling, preceding = NULL) {
```
Melkiades, line 274: `at_sibling` reaches here unvalidated, so the bad-input shapes surface internals:

| input | message |
|---|---|
| `at_sibling = 1` | `Unable to find structural element '1'` |
| `at_sibling = c("SEX", "AGE")` | `values must be length 1, but FUN(X[[1]]) result is length 2` |
| `at_sibling = "SEX[0]"` | `argument is of length zero` |

A `checkmate::assert_string(at_sibling, null.ok = TRUE)` at the entry points plus a lower bound on the index in `extract_dup_pos()` covers all three. On a fresh layout the message also comes out as `Eligible elements: ''`.

Tiny one while you are here: the `2` suffix, there is no `find_branch_pos` left in the package.

R/nesting_impl.R
```diff
@@ -0,0 +1,621 @@
...
+#' @rdname int_methods
+setMethod(
+  "split_rows", "PreDataRowLayout",
+  function(lyt, spl, pos, cmpnd_fun = AnalyzeMultiVars, at_sibling = NULL) {
+    stopifnot(is.na(pos) || (pos > 0 && pos <= length(lyt) + 1))
+    root_branching <- FALSE
+    if (!is.null(at_sibling)) {
+      oldval <- lyt[[pos]]
+      ## if we at_sibling a top level element we need to handle as nested = FALSE
+      if (branch_is_root(oldval, at_sibling)) {
```
Melkiades, line 376: Two things here.

The `has_force_pag()` check below reads `last_rowsplit(oldval)`, the last split in the chain, rather than the split at the anchor, so it only fires when those coincide. This errors as intended:

```r
basic_table() |> split_rows_by("STRATA1", page_by = TRUE) |>
  split_rows_by("SEX", at_sibling = "STRATA1")
#> Error: at_sibling pointed to a split with forced pagination ...
```

but insert one `analyze("AGE")` before the `at_sibling` line and the same anchor builds quietly.

Also `at_sibling` together with `nested = FALSE` gives `subscript out of bounds`. They conflict, so an explicit message would land better than the subscript error.

R/tt_showmethods.R
```diff
@@ -41,6 +41,13 @@ setMethod(
   }
 )
 
+setMethod(
+  "ploads_to_str", "SplitVectorTree",
```
Melkiades, line 45: Printing a layout that uses the feature flattens the branch and leaks a literal `c(...)`:

```
Row-Split Structure:
STRATA1 (lvls) -> c("SEX", "AGE", "RACE", "AGE") ()
```

This returns a vector and the `SplitVector` show method then deparses it, so the `docat_splitvec` tree branch further down never gets reached for this shape. Given the branch structure is the point of the feature, seeing it in `print(lyt)` would help a lot.

R/tree_accessors.R
```diff
@@ -753,6 +653,11 @@ setMethod("label_position", "Split", function(spl) spl@split_label_position)
 #' @rdname int_methods
 setMethod("label_position", "VAnalyzeSplit", function(spl) spl@var_label_position) ## split_label_position)
 
+#' @rdname int_methods
+setMethod("label_position", "SplitVectorTree", function(spl) {
```
Melkiades, line 657: Getter is here, but `label_position<-` still only has a `Split` method, and `branch_above_split()` does `label_position(splvec[[branch_pos]]) <- "visible"` where that element can be a tree.

I could not actually reach it, the `endontree && !sib_matches` combination looks unreachable and your comment there says the same. So either the setter or a `stopifnot` making the invariant explicit, whichever you prefer.

R/00tabletrees.R
```diff
@@ -1914,6 +1913,10 @@ TableTree <- function(kids = list(),
 ## a pre-existing TableTree/ElementaryTable.
 ## This is used for add_existing_table in colby_constructors.R
 
+split_or_splitvectree <- function(object) {
```
Melkiades, line 1916: `head(object, -1)` means a `SplitVectorTree` now validates at any position, but the implementation assumes trees are terminal, `branch_above_split()` has `stopifnot(is(splvec[[svlen]], "SplitVectorTree"))` on exactly that. Restricting the relaxation to the last element would keep the validity function saying what the code relies on.

R/tt_dotabulation.R
```diff
@@ -1452,7 +1521,7 @@ build_table <- function(lyt, df,
     if (length(splvec) == 0) {
       return(NULL)
     }
-    firstspl <- splvec[[1]]
+    firstspl <- unlist(splvec, recursive = TRUE)[[1]] ## could be a SplitVecTree now...
```
Melkiades, line 1524: This idiom, `unlist(..., recursive = TRUE)[[1]]` to reach the first split, now appears in `first_spl_name()`, `.make_split_kids`, `recursive_applysplit` (twice) and here, in slightly different spellings each time. A small `first_split()` helper would collapse all five and make the `## could be a SplitVecTree now...` note unnecessary.

Line 1554 also still has the commented `## kids <- kids[!sapply(kids, is.null)]` that `unlist()` replaced.

R/nesting_impl.R
```diff
@@ -0,0 +1,621 @@
...
+  sib_matches <- is.null(at_sibling) || deuniqify_path_elements(at_sibling) %in% vapply(lstlastel, first_spl_name, "")
+  if (endontree && sib_matches) {
+    splvec[[branch_pos]] <- SplitVectorTree(lst = c(lastel, list(SplitVector(newspl))))
+  } else if (has_force_pag(lastel)) {
+    stop(
+      "at_sibling pointed to a split with forced pagination (page_by = TRUE).",
+      " This is not supported."
+    )
+  } else {
+    ## are_spls <- which(!vapply(splvec, is, "VAnalyzeSplit", FUN.VALUE = TRUE))
```
Melkiades, line 350: Three leftovers around here:

* these two commented lines survived the cleanup commit
* `branch_pos > 0` on the next line is always true, `find_branch_pos2()` returns a `which()` result
* `is.null(at_sibling)` in `sib_matches` above is always `FALSE` on this path, since `branch_above_split()` is only called when an anchor was given

And at the signature, `branch_pos`'s default references `preceding`, which is declared after it. Lazy evaluation makes it work, it just reads as a bug on the way past.

R/00tabletrees.R
```diff
@@ -466,7 +466,7 @@ MultiVarSplit <- function(vars,
                           colcount_format = NULL) {
   check_ok_label(split_label)
   ## no topleft allowed
-  label_pos <- match.arg(label_pos, label_pos_values[-3])
+  label_pos <- match.arg(label_pos, label_pos_values[-4])
```
Melkiades, line 469: This had to move from `[-3]` to `[-4]` because an element was prepended to `label_pos_values`. `setdiff(label_pos_values, "topleft")` would not need touching next time.

Same shape at lines 725 and 779, where `c("default", label_pos_values)` now carries a duplicate since `"default"` is in the vector. `match.arg` resolves it fine, I checked, it is just leftover scaffolding.

NEWS.md
```diff
@@ -20,6 +20,7 @@
  * Added vignette: Guided Tour (Advanced) - (stub) `TableTree` Objects - (stub) Accessing Table Values @gmbecker
  * Added vignette: Guided Tour (Advanced) - (stub) `TableTree` Objects - (stub) Custom Scoring Functions For Sorting @gmbecker
  * Added vignette: Guided Tour (Advanced) - (stub) `TableTree` Objects - (stub) Custom Pruning Functions @gmbecker
+ * Added Intermediate nesting behavior @gmbecker
```
Melkiades: One line for the whole feature. `at_sibling`, `get_anchor_list()` and the new `label_pos = "default"` are all user-visible and none are mentioned.

The `call. = FALSE` removal in `uniqify_child_names` deserves its own Bug Fixes entry too. `message()` has no `call.` argument so it was being pasted into the text and users were seeing it end in a literal `FALSE`. Good catch, it just gets lost in here.

Trailing whitespace on this line.

R/argument_conventions.R
```diff
@@ -150,6 +150,10 @@ gen_args <- function(df, alt_counts_df, spl, pos, tt, tr, verbose, colwidths, ob
 #'   to the *split* or *group of sibling analyses*, for `split_rows_by*` and
 #' `analyze*` when analyzing more than one variable, respectively. Ignored when
 #' analyzing a single variable.
+#' @param at_sibling (`character(1)` or `NULL`)\cr If non-null, a preceding
```
Melkiades, line 153: Reads like it lost a couple of words: "downstream of a point a previously used anchor".

R/colby_constructors.R
```diff
@@ -438,6 +190,44 @@ setMethod(
 #'
 #' @inheritSection custom_split_funs Custom Splitting Function Details
 #'
+#' @section Nesting Anchor Resolution:
+#'
+#' When `nested` is `TRUE`, `at_sibling` allows you to set a *nesting
+#' anchor* that your new `split_rows_by*` or `analyze*` directive
+#' should be placed as a sibling to.  The lookup for this anchor
+#' occurs *only in the currently active top-level nesting stack*,
+#' meaning the directives splitting directives that have occurred since
```
Melkiades: Two in this section: "the directives splitting directives that have occurred" here, and at line 225 "It is strongly suggested that the `parent_name` and `table_names` argument(s) of `split_rows_by*` and `analyze` to prevent the need for this" is missing its verb.

The section itself is good. I checked the masking behaviour it describes and it matches what the code actually does.

tests/testthat/test-nesting.R
```diff
@@ -0,0 +1,553 @@
...
+    split_rows_by_cutfun("AGE", at_sibling = "FACTOR2") |>
+    analyze("AGE")
+  tbl <- build_table(lyt, rawdat)
+
+  expect_gt(path_count(tbl, c("RACE", "WHITE", "AGE", "1st qrtile", "AGE", "Mean")), 0L)
+  expect_true("AGE" %in% row.names(tbl))
+})
+
+
+test_that("intermediate nesting works correctly", {
```
Melkiades, line 106: This block runs about 450 lines and 40 assertions, so a failure in the first few hides the rest. Splitting it would pay off the first time something regresses.

On coverage, the shapes that bit me are not in here: a second consecutive `analyze()` after a branch (the blocker), `vars_in_layout()` on a branched layout, and `get_anchor_list()` output fed back into `at_sibling`.

vignettes/guided_intro_nesting.Rmd
```diff
@@ -59,31 +68,384 @@ library(rtables)
...
+1. `analyze`ing `AGE`, then
+2. splitting by `SEX` and `analyze`ing `AGE`, and finally
+3. splitting by `RACE` and `analyze`ing `AGE` again.
+
+We will start each section after the first with `nested = FALSE` to
+delineate it from the previous portion of the layout.
+
+NOTE: while we will do it explicitly for illustration purposes, any
+`split_rows_by` layout instruction that follows an `analyze` is
+defaults to `nested = FALSE`.
```
Melkiades, line 114: Typo: "that follows an `analyze` is defaults to `nested = FALSE`".

Melkiades (changes requested):
Really like where this lands. I checked out both branches and built 20 standard layouts (nested, `nested = FALSE`, all four `label_pos` values, cuts, quartiles, cutfun, multivar, `analyze_colvars`, `summarize_row_groups`, `page_by`, `add_overall_col`, topleft, `vars_in_layout`, layout printing) on `main` and on this head: 20/20 byte identical. The `label_pos = "default"` addition does not move any existing output, which was the thing I was most worried about going in. `summarize_row_groups`, sorting, pagination, `split_rows_by_multivar`, three-way branching on one anchor and `parent_name` disambiguation all work on branched layouts too.

Worth calling out separately: dropping `call. = FALSE` from the `uniqify_child_names` `message()` is a real fix. `message()` has no `call.` argument so it was being pasted in, and users were seeing that message end in a literal `FALSE`. That deserves its own NEWS bug-fix line rather than riding along in here.

One correctness issue blocks, then a set of smaller things. Most of the smaller ones are the new exported surface not yet covering the structure the feature creates.

**Blocking**

1. A second consecutive bare `analyze()` after an `at_sibling` branch collapses the split structure and computes on unsplit data, silently. Reproducer inline on `cmpnd_last_rowsplit`.
2. `vars_in_layout()` errors on any branched layout. Inline on the class definition, with the other generics in the same position.
3. `get_anchor_list()` lists anchors that `at_sibling` then rejects.

**Worth folding in**

4. `at_sibling` takes no input validation: numeric, length > 1 and `"SEX[0]"` each surface a raw internal error.
5. `at_sibling` with `nested = FALSE` gives `subscript out of bounds`.
6. `print(lyt)` on a branched layout leaks a literal `c("SEX", "AGE", "RACE", "AGE")` and hides the branch.
7. The `page_by` guard tests the last split in the chain rather than the split at the anchor.
8. `label_position<-` has no `SplitVectorTree` method although `branch_above_split()` can reach for one.
9. `SplitVector` validity now allows a tree at any position while the code assumes terminal.
10. NEWS is one line for the whole feature.

**Polish**, all inline and all take-or-leave: a repeated `unlist(...)[[1]]` idiom worth a helper, some commented-out code that survived the cleanup commit, `find_branch_pos2` naming, positional indexing into `label_pos_values`, one very long `test_that` block, four typos in the new prose.

Style Check is the one red job.

Happy to move any of the polish items to a follow-up issue if you would rather keep this PR tight.

vignettes/guided_intro_nesting.Rmd
```diff
@@ -59,31 +68,384 @@ library(rtables)
...
+
+1. `analyze`ing `AGE`, then
+2. splitting by `SEX` and `analyze`ing `AGE`, and finally
+3. splitting by `RACE` and `analyze`ing `AGE` again.
+
+We will start each section after the first with `nested = FALSE` to
+delineate it from the previous portion of the layout.
+
+NOTE: while we will do it explicitly for illustration purposes, any
+`split_rows_by` layout instruction that follows an `analyze` is
```
munoztd0:
```suggestion
`split_rows_by` layout instruction that follows an `analyze` 
```

R/00tabletrees.R
gmbecker: Fixed:

```
> keep_2_levels <- function(varnm, dat = ex_adsl) keep_split_levels(levels(dat[[varnm]])[1:2])
> clown_nose <- basic_table() |>
+ split_rows_by("STRATA1", split_fun = keep_2_levels("STRATA1")) |>
+ split_rows_by("STRATA2", split_fun = keep_2_levels("STRATA2")) |>
+ analyze("ARM") |>
+ split_rows_by("SEX", split_fun = keep_2_levels("SEX")) |>
+ split_rows_by("RACE") |>
+ split_rows_by("STRATA1", split_fun = keep_2_levels("STRATA1")) |>
+ analyze("BMRKR1") |>
+ split_rows_by("BMRKR2", split_fun = keep_2_levels("BMRKR2"), at_sibling = "RACE") |>
+ split_rows_by("COUNTRY", split_fun = keep_2_levels("COUNTRY")) |>
+ analyze("AGE") |>
+ split_rows_by("SITEID", split_fun = drop_split_levels, at_sibling = "RACE") |>
+ split_rows_by("BEP01FL") |>
+ analyze("AGE")
> vars_in_layout(clown_nose)
 [1] "STRATA1" "STRATA2" "ARM"     "SEX"     "RACE"    "BMRKR1"  "BMRKR2"  "COUNTRY" "AGE"     "SITEID"  "BEP01FL"
```

R/nesting_impl.R
gmbecker: fixed

```
> basic_table()|> split_rows_by("STRATA1", page_by = TRUE) |> analyze("AGE") |> split_rows_by("SEX", at_sibling = "STRATA1")
Error in find_branch_pos_df(anchordf = anchordf, at_sibling = at_sibling) (from nesting_impl.R#597) : 
  at_sibling pointed to an element with forced pagination (page_by = TRUE). This is not supported.
```
and 
```
> lyt <- basic_table() |>
+ split_rows_by("SEX") |>
+ split_rows_by("STRATA1") |>
+ analyze("AGE") |>
+ split_rows_by("SEX", nested = FALSE, at_sibling = "STRATA1")
```

R/nesting_impl.R
gmbecker: fixed (with the caveat that the accessor is now called `get_row_anchor_list` which relies on completely refactored code internally):
```
> lyt <- basic_table() |>
+ split_rows_by("SEX") |>
split_rows_by("STRATA1") |>
+ + analyze("AGE") |>
+ split_rows_by("SEX", nested = FALSE, at_sibling = "STRATA1")
> lyt <- basic_table() |>
+ split_rows_by("SEX") |> analyze("AGE") |>
+ split_rows_by("BMRKR2", nested = FALSE) |> split_rows_by("RACE") |> analyze("AGE")
> get_row_anchor_list(lyt)
[[1]]
[1] "SEX"

[[2]]
[1] "BMRKR2"

[[3]]
[1] "RACE"

[[4]]
[1] "AGE"

> lyt |> split_rows_by("STRATA1", at_sibling = "SEX")
A Pre-data Table Layout

Column-Split Structure:
<implicit> (all obs)

Row-Split Structure:
SEX (lvls)     -> AGE (** var **)                   
BMRKR2 (lvls)  -> RACE (lvls)     -> AGE (** var **)
STRATA1 (lvls)                                      

'->' indicates nesting, vertical stacks of '|' indicate anchoring/siblings.
'(<type>)' indicates split type, while '(** <type> **)' indicates an analyze instruction.
```

I'm going to leave the nesting (its not really nesting, simply that elements of the anchor list can have length >1 indicating an anchor point and its siblings) as I think its useful and you can do `get_row_anchor_df` to get a data.frame now where the name column is the anchor point names as a vector

R/nesting_impl.R
gmbecker: Good catch, fixed:

```
   lyt <- basic_table() |>
        split_rows_by("STRATA1", split_fun = keep_2_levels("STRATA1")) |>
        split_rows_by("SEX", split_fun = keep_2_levels("SEX")) |>
        analyze("AGE", table_names = "a1") |>
        split_rows_by("RACE", split_fun = keep_2_levels("RACE"),
                      at_sibling = "SEX") |>
        analyze("AGE", table_names = "a2") |>
        analyze("BMRKR1")
```

```
> lyt
A Pre-data Table Layout

Column-Split Structure:
<implicit> (all obs)

Row-Split Structure:
STRATA1 (lvls) -> | SEX (lvls)  -> a1 (** var **)               
                  | RACE (lvls) -> ma_a2_BMRKR1 (** multivar **)

'->' indicates nesting, vertical stacks of '|' indicate anchoring/siblings.
'(<type>)' indicates split type, while '(** <type> **)' indicates an analyze instruction.
```

tests/testthat/test-nesting.R
gmbecker: fixed

R/nesting_impl.R
gmbecker: added, and find_branch_pos(2) is no longer there.

gmbecker:
All comments resolved.

gmbecker:
clean scda.test https://github.com/insightsengineering/scda.test/pull/255 and ready to merge @shajoezhu

PR #1122, opened 2026-09-21
adding validation reports

shajoezhu:
**What changes are proposed in this pull request?**

* Adds a "Validation report" entry to the pkgdown navbar, resolved per documentation version. (#1122)

pkgdown builds one static site per version, so a hardcoded link would freeze at whichever report was current when it was written. `pkgdown/extra.js` reads the version from the navbar and points the link at the newest report at or below it. Doc versions older than the first report drop the item entirely. `_pkgdown.yml` keeps a static href as the fallback for readers with JavaScript disabled.

Only 0.6.16 currently has a report, so 0.6.16 and later resolve to it and earlier doc versions get no navbar item.

Same change as pharmaverse/formatters#377.


Melkiades:
@shajoezhu Same three fixes here as on pharmaverse/formatters#377, pushed as 2b3e3810, plus a description since the body was empty.

* `findLink()` scanned every anchor on the page, so a vignette linking the words "validation report" would have been rewritten instead of the navbar item. Scoped to the navbar.
* `closest()` and `remove()` were the only ES6-era DOM calls in an otherwise strict ES5 file. Swapped for equivalents.
* `pickReport()` took the last match, so it depended on `REPORTS` being sorted. It now takes the maximum. The header comment also said adding a version needs "no other change", but the static href in `_pkgdown.yml` that no-JS readers get has to move too.

Checked for rtables specifically: the 0.6.16 report is a 200 while 0.6.15 and 0.6.17 are 404, so the single-entry `REPORTS` is right; `main` at 0.6.16.9007 resolves to 0.6.16; and the live nesttemplate navbar's first `<small>` is `0.6.16`, so `detectVersion()` finds it. Resolution across versions: 0.6.14 and 0.6.15 drop the link, 0.6.16 onwards get 0.6.16.

One left for you, same as on the formatters PR: an rc such as `0.6.16-rc1` would resolve to the 0.6.16 report, since `parseInt("rc1")` is `NaN` and becomes 0 so the rc compares equal to the release. Happy to change it if you would rather rc builds got the previous report.

shajoezhu:
i think the release candidate tags, we can remove them

Melkiades (approved):
Lgtm! Thanks Joe

